PR #1, opened 2025-05-20
chore: project setup

HendrikThePendric:
# Description
This branch sets up the project

## Todo

- [x] Initialize new DHIS2 app with Typescript, using CLI App Scripts
- [x] Configure various linters and formatters
    - [x] ESLint
    - [x] ls-lint
    - [x] Prettier
    - [x] Stylelint
- [x] Unit testing using vitest
- [x] E2E testing using cypress
- [x] Git Hooks (pre-commit + pre-push)
- [x] GitHub workflows (limited to the ones we need right now), fix the following:
    - [x] Recording E2E runs
    - [x] Deploying to Netlify
    - [x] Running the BOM
- [x] Pull request template
- [x] Dependabot settings
- [x] Readme
- [x] Types

### DIFF
--- a/cypress/plugins/exclude-by-version-tags.ts
+++ b/cypress/plugins/exclude-by-version-tags.ts
@@ -11,7 +11,7 @@ untagged and be run against all supported versions of DHIS2.
 
 DHIS2 officially supports the latest 3 released versions of DHIS2. For example: 2.40,
 2.41 and 2.42. Dev would then have version 2.41-SNAPSHOT. Therefore, the getExcludedTags
-alculates the range of tags based on minimum supported version + 3 (2.40, 2.1, 2.42, 2.43-SNAPSHOT)
+calculates the range of tags based on minimum supported version + 3 (2.40, 2.41, 2.42, 2.43-SNAPSHOT)
 With the minimum supported version of 2.40, the tags will always contain "40", "41" and "42",
 but the comparison symbols will depend on the current instance version.
 Allowed tag comparisons are ">", ">=", "<", "<="


.github/workflows/check-tasklist.yml
```diff
@@ -0,0 +1,17 @@
...
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+jobs:
+    validate-checklists:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - name: Validate Checklists
+              uses: roryq/checkmate@master
```
HendrikThePendric, line 17: New action

.github/workflows/verify-pr.yml
```diff
@@ -0,0 +1,119 @@
...
+        types: ['opened', 'labeled', 'reopened', 'synchronize']
+
+env:
+    GIT_AUTHOR_NAME: '@dhis2-bot'
+    GIT_AUTHOR_EMAIL: 'apps@dhis2.org'
+    GIT_COMMITTER_NAME: '@dhis2-bot'
+    GIT_COMMITTER_EMAIL: 'apps@dhis2.org'
+    GH_TOKEN: ${{secrets.DHIS2_BOT_GITHUB_TOKEN}}
+    CI: true
+    DEV_URL: 'https://test.e2e.dhis2.org/anly-e2e-dev'
```
HendrikThePendric, line 14: We're only testing against dev, because this is a new app.

.ls-lint.yml
```diff
@@ -0,0 +1,15 @@
...
+        .dir: kebab-case
+        .ts: kebab-case
+        .test.ts: kebab-case
+        .tsx: kebab-case
+        .test.tsx: kebab-case
+        .d.ts: kebab-case
+        .cy.ts: kebab-case
+        .module.css: kebab-case
+        .sh: kebab-case
+        .json: kebab-case
```
HendrikThePendric, line 12: IMO all files and folders should use the same casing. I like kebab-case, but happy to change to something else, as long as it's consistent

.ls-lint.yml
```diff
@@ -0,0 +1,15 @@
...
+
+ignore:
+    - types/dhis2-openapi-schemas
```
HendrikThePendric, line 15: These generated types are using `CamelCase`

scripts/generate-test-matrix.cjs
HendrikThePendric, line 1: I thought moving this into the `./scripts` dir made sense. We have it in `./cypress/support` in our other apps, but even though this script is used to distribute the Cypress test files across runners on CI, it's not technically _part of_ the Cypress setup.

vitest.setup.ts
```diff
@@ -0,0 +1,9 @@
+import * as matchers from '@testing-library/jest-dom/matchers'
+import { cleanup } from '@testing-library/react'
```
HendrikThePendric, line 2: So most of the actual test code is actually going to be using the React Testing Library, which is what I believe we've settled on as an organisation.

.github/pull_request_template.md
```diff
@@ -0,0 +1,39 @@
...
+
+---
+
+### ToDos
+
+<!--Checkmate-->
+
+-   [ ] _todo_
+
+---
```
HendrikThePendric, line 29: Some minor changes for the current approach here. I've switched to a different GHA to check our checklists, which actually fails the check if items are left unchecked. I much prefer that from a visual perspective. This action only validates checklists with a  comment above them, so I've added those to the template.

This action does not pick up on N/A strings in the item, so it will require a slightly different approach. Instead of adding N/A and leaving the box unchecked, we can add N/A and then check the checkbox.

.github/pull_request_template.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Known issues
+
+-   _issue_
+
+---
```
HendrikThePendric, line 35: Perhaps a bit pedantic, but I don't think "Known issues" should be checkboxes. When would you actually check the box? I've changed them to regular bullet points for now, because I think that's more apt, but happy to discuss of course.

.github/workflows/dhis2-verify-commits.yml
```diff
@@ -0,0 +1,44 @@
...
+        types: ['opened', 'edited', 'reopened', 'synchronize']
+
+jobs:
+    lint-pr-title:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: 18.x
```
edoardo, line 14: Should we start using 20.x?
GHA transitioned all actions to Node v20 already in spring 2024.

cypress/plugins/exclude-by-version-tags.ts
```diff
@@ -0,0 +1,109 @@
...
+The list of excluded tags returned by getExcludedTags are the tags that cypress will
+ignore when running the test suite. So if a test is tagged with one of the tags in the
+excluded list, then that test will not run.
+
+Using excluded tags (instead of included tags) allows for most of the tests to remain
+untagged and be run against all supported versions of DHIS2.
+
+DHIS2 officially supports the latest 3 released versions of DHIS2. For example: 2.40,
+2.41 and 2.42. Dev would then have version 2.41-SNAPSHOT. Therefore, the getExcludedTags
+alculates the range of tags based on minimum supported version + 3 (2.40, 2.1, 2.42, 2.43-SNAPSHOT)
```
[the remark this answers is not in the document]
jenniferarnesen: also, 2.41, not 2.1

edoardo (approved):

PR #12, opened 2025-07-29
feat: implement VisualizationTypeSelector component (DHIS2-19819)

edoardo:
Implements [DHIS2-19819](https://dhis2.atlassian.net/browse/DHIS2-19819)

### Description

Implements the `VisualizationTypeSelector` component similarly to the one used in DV.
The 2 available vis types for now are `Line listing` and `Pivot table`.
The component is connected to the Redux store and reads/writes `visualizationType` in the `ui` slice of the store to keep the current vis type in sync.

---

### Quality checklist

- [x] Dashboard tested _N/A_
- [x] Docs added _N/A_
- [x] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/XXX>) _N/A_

---

### ToDos

- [x] connect the component for storing the selected vis type
- [x] double check which vis type should be selected by default

---

### Known issues

-   The `IconVisualizationLinelist24` and `IconVisualizationPivotTable24` from `ui` work fine in the closed selector, although in DV the icons there have colors. Because it's not possible to specify a size for the icons from `ui`, the big icons in the dropdown of the selector need to be custom made specifically for the app. In DV we use custom icons for both the closed selector (small icon) and the dropdown (big icon).
There is a separate ticket for this and needs some input from the Design team:
https://dhis2.atlassian.net/browse/DHIS2-19961

---

### Screenshots
Screenshot 2025-07-30 at 09 55 01

[DHIS2-19819]: https://dhis2.atlassian.net/browse/DHIS2-19819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/toolbar/visualization-type-selector/visualization-type-list-item.tsx
```diff
@@ -0,0 +1,42 @@
+import cx from 'classnames'
+import React, { FC } from 'react'
+import { ListItemIcon } from './list-item-icon'
+import classes from './styles/visualization-type-selector.module.css'
+
+type VisualizationTypeListItemProps = {
+    description: string
+    iconType: string
```
edoardo, line 8: This should actually use the type for the visualization types.

src/components/toolbar/visualization-type-selector/visualization-type-list-item.tsx
```diff
@@ -0,0 +1,42 @@
...
+    isSelected: boolean
+    label: string
+    onClick: () => void
```
edoardo, line 11: This is an event so it needs the proper type.

src/modules/visualization.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const filterVisTypesByVersion = (visType: SupportedVisType) =>
+        // only PT and LL enabled in the first version
+        (serverVersion?.minor ?? 0) < 44 &&
+        SUPPORTED_VIS_TYPES.includes(visType)
+            ? true
+            : false
+
+    return filterVisTypesByVersion
+}
```
HendrikThePendric, line 28: We discussed this a bit already. The hook returns a function and that's OK, but you could consider rewriting to something like this.

```typescript
export const useVersionFilteredVisTypes = (): Array<SupportedVisType> => {
    const { serverVersion } = useConfig()
    
    if (typeof serverVersion?.minor !== 'number') {
        throw new Error('Server version minor was not a number')
    }
    
    // only PT and LL enabled in the first version
    return SUPPORTED_VIS_TYPES.filter((visType) =>
        serverVersion.minor > 43 || visType === 'LINE_LIST' || visType === 'PIVOT_TABLE'
    )
}
```

This returns an array you can iterate over directly. The hook doesn't require any arguments. So at first glance this seems more ergonomic. But I don't know exactly what other cases we would use this for. So perhaps having a hook that returns a function is what we need....

Having said that, for these cases we could also have another helper like this:
 ```typescript
 const isVisTypeSupportedOnVersion = (visType: string, versionMinor: number) =>
    versionMinor > 43 || visType === 'LINE_LIST' || visType === 'PIVOT_TABLE' 
 ```

And we could even use the helper in the hook.

HendrikThePendric (commented):
I know you're still busy with this, but I left a few comments.

Also not that some of the CSS and markup in `app.module.css` may be (re)moved when I work on the app skeleton.

src/types/analytics/index.d.ts
```diff
@@ -1,9 +1,15 @@
...
 } from './cached-data-query-provider'
+import type { Toolbar } from './toolbar'
+import type { ToolbarSidebar } from './toolbar-sidebar'
 
 declare module '@dhis2/analytics' {
     export const CachedDataQueryProvider: CachedDataQueryProvider
     export const useCachedDataQuery: useCachedDataQuery
+    export const Toolbar: Toolbar
+    export const ToolbarSidebar: ToolbarSidebar
+    export const visTypeDisplayNames: Array<Record<SupportedVisType, string>>
```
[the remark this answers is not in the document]
edoardo: Changed.

PR #25, opened 2025-08-12
chore: add cypress component tests and address TS Config and lint issues

HendrikThePendric:
ISSUE N/A

### Description

When I started this branch I was planning on implementing issue DHIS2-19816, the app-layout components. While I did add a few components here, I actually ended up implementing various different things:
- We can now use Cypress Component tests, these are useful in case you want to test a component in isolation (so no E2E test), but you need a real DOM (computed styles, scroll bars, etc don't work in jsDOM).
- There were a few issues with TSConfigs and as a result the Cypress global types would often not be picked up by VSCode and/or the TS compiler. ~This has now been addressed by reorganising and tweaking the TS Config files (we now have `tsconfig.json`, `tsconfig.cypress-e2e.json` and `tsconfig.cypress-component.json`.~ This had to be resolved in quite weird way. I'll provide details below.
- And when all that was done, I realised that `@dhis2/cli-style` has various issues which resulted in files not being linted/formatted properly. After trying a lot of different approaches I got quite frustrated and ultimately decided to completely stop using the CLI Style commands. We are still using the config files from the CLI style lib, so the code we will be producing is still going to be compliant in terms of structure and syntax, but we're just using the various tools directly. I produced a few bash files that do that for us. In general, the approach I took is to not exit early: all checks will complete even if the first one fails. This is slower but it enables you to see all issues in one go.

#### TSConfig files

The most logical solution for a project that needs different TS Configs for different files and folders is to work with a several `composite` TS Configs which are then referenced in a main `tsconfig.json`. This was the solution I implemented initially, and while this does work from a TS compiler perspective, it turns out editors (neither NeoVIM or VSCode) cannot work with this type of setup reliably. Sometimes it works, but after a while the editor "gets confused" and starts throwing TS errors, which makes it pretty much impossible to work in the project. 

So I had to look for an alternative approach, which turned out to be extremely complicated. The were some solutions that worked for editor, but broke the linting in the terminal (and on CI). Other solutions ended breaking the Cypress run/open commands, for reasons I still don't fully understand. In the end, the solution I found to work also came with a set of problems, which could be worked around gracefully:
- We have two TS Config files, both called `tsconfig.json`, one at the project root which is for everything not in the `cypress` dir, and one `cypress/tsconfig.json` which is for the Cypress E2E code. Working with these nested config files with a default name was something the TS compiler in the editors and terminal could deal with well.
- The main TSConfig file exposes global Cypress types to our files in src, which is problematic for the spec files, because these were also using global types with the same name (`expect`, `describe`, etc).
- In Vitest, you can disable globals, which means everything has to be imported like a normal variable, and for the TS compiler these imported variable types take precedence over the globally defined ones. So that's a solution for these overlapping types. So I disabled `global` in the Vitest config and adjusted the spec files, so they imported things like `expect`, `it`, `describe`, etc.
* But now there was still another issue to solve: when writing a new spec file the tests will fail without importing `expect` etc, but the editor does not pick up on it, because it's also a global Cypress variable. So this is confusing, but I managed to solve it with an ESLint rule. I added a `no-restricted-globals` override for the file glob `'src/**/*.spec.{ts,tsx}'` and added all the Cypress globals that overlap with Vitest. So now when you forget to import `describe` etc in a spec file, the editor will flag it up too

#### AppLayout components
I considered removing all the app-layout components from this PR, then I would have to adjust my Cypress Component test. I think at this stage of development it's OK to just merge them in alongside the project-setup related code.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A
- [x] Tester approved (JJA)

---

### Screenshots

_New lint output_
Screenshot 2025-08-14 at 17 35 48

_Cypress component tests (UI)_
Screenshot 2025-08-14 at 17 38 02

_Cypress component tests (CI)_
Screenshot 2025-08-14 at 17 40 44

_Cypress types_
Screenshot 2025-08-14 at 17 38 48


scripts/format.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
jenniferarnesen, line 1: Are these scripts written from scratch or copied and modified? If they were copied, it would be useful to include in the comment where it was copied from.

Hopefully in the long term we can move back to d2-style, but I understand the need to do this now to be able to move forward.

src/components/skeleton/skeleton-canvas.module.css
```diff
@@ -0,0 +1,6 @@
+.container {
```
jenniferarnesen, line 1: Are we intentionally moving away from having a separate styles folder? I admit I liked having the folder because it was easier to get an overview of the components in the folder. But if this is the new convention, then I'm fine with it too.

src/components/skeleton/skeleton-canvas.tsx
```diff
@@ -0,0 +1,8 @@
+import type { FC, ReactNode } from 'react'
```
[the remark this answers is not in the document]
HendrikThePendric: As discussed, I changed the name to `AppLayout`

jenniferarnesen (approved):

PR #28, opened 2025-08-22
feat: app layout grid [DHIS2-19816]

HendrikThePendric:
Implements [DHIS2-19816](https://dhis2.atlassian.net/browse/DHIS2-19816)

### Description

Adds app layout components using CSS Grid model. The original layout in LL is implemented using the flexbox model, but it turned out that CSS Grid was actually much more suitable for trying to achieve our desired layout and as a result we needed fewer elements/components. We also decided that instead of using "skeleton" we would use "grid" in the component names. So the final layout is quite a bit different than the one depicted in the issue description. The current ones are as below:

AppGrid

---

### Quality checklist

-   [x] Dashboard tested N/A
-   [x] Cypress Component Tests added
-   [x] Docs added  N/A
-   [x] d2-ci dependency replaced N/A
-   [x] Tester approved (@edoardo)

---


src/components/grid/__tests__/grid.cy.tsx
```diff
@@ -0,0 +1,438 @@
...
+            FULL_HEIGHT - defaultCenterColumnTopContentSize.height - 20
+        )
+        // Top row content grows to specified size with overflow hidden
+        assertHeight('top-row-content', excessiveHeight)
+        // Center column top remains the specified height
+        assertHeight(
+            'grid-center-column-top',
+            defaultCenterColumnTopContentSize.height
+        )
+        // Center column bottom shirnks to the value of grid-template-rows bottom row minmax value
```
edoardo, line 261:
```suggestion
        // Center column bottom shrinks to the value of grid-template-rows bottom row minmax value
```

edoardo (dismissed):

PR #29, opened 2025-08-22
feat: implement Interpretations and detail toggler in toolbar

edoardo:
Part of [DHIS2-19821](https://dhis2.atlassian.net/browse/DHIS2-19821)

### Description

Adds the Interpretations and details toggler button to the toolbar.
The toggler is connected to the store using the same key in the `ui` slice as in LL app.
The toggler changes both `showDetailsPanel` and `showAccessoryPanel` to keep the same behaviour as in the current LL app.

Note: both the details panel and the accessory panel are not implemented yet.
Cypress tests will be added once those are available.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

Note: the button will show in the correct place to the far right of the toolbar once the MenuBar component is added.
Screenshot 2025-08-22 at 15 59 12

[DHIS2-19821]: https://dhis2.atlassian.net/browse/DHIS2-19821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/store/current-slice.ts
+++ b/src/store/current-slice.ts
@@ -1,0 +1,25 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { EventVisualization } from '@types'
+
+type CurrentState = Partial<EventVisualization> | null
+
+const initialState: CurrentState = null
+
+export const currentSlice = createSlice({
+    name: 'current',
+    initialState,
+    reducers: {
+        setCurrent: (
+            state: CurrentState,
+            action: PayloadAction<CurrentState>
+        ) => {
+            state = action.payload
+        },
+    },
+    selectors: {
+        getCurrent: (state: CurrentState) => state ?? null,
+        getCurrentId: (state: CurrentState) => state?.id ?? null,
+    },
+})
+
+export const { setCurrent } = currentSlice.actions
--- a/src/store/current-slice.ts
+++ b/src/store/current-slice.ts
@@ -1,7 +1,7 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit'
 import { EventVisualization } from '@types'
 
-type CurrentState = Partial<EventVisualization> | null
+type CurrentState = EventVisualization | null
 
 const initialState: CurrentState = null
 
@@ -23,3 +23,4 @@ export const currentSlice = createSlice({
 })
 
 export const { setCurrent } = currentSlice.actions
+export const { getCurrent, getCurrentId } = currentSlice.selectors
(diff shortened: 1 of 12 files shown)


src/types/analytics/index.d.ts
edoardo, line 1: @HendrikThePendric is this supposed to be renamed to `index.ts`?

HendrikThePendric: I'm not quite sure actually. In general file I'd say regular `.ts` file are preferable over `d.ts` file because these make the types globally available in the project. However, since this file is acting as the type definition file for the lib, it could well be that things break if we change it into a `.ts` file.

Taken the above into account, I'd say:
Try to change it to `.ts` and if the TS compiler still is able to pick up the types we've added there from the imports from `@dhis/analytics`, then we stick with it.

HendrikThePendric (changes requested):
This code is looking great @edoardo and I greatly appreciate that you took the time to split up your work in smaller batches. With the PR being relatively small I felt like I was able to review it very well.

I've left a few minor comment and an answer to your question. Perhaps it's still a bit early to say, but I can perhaps see one general pattern emerging already, which is maybe not ideal:
- In order to use the selectors, we first destructure the state-slice, because otherwise we end up accessing a property three levels deep `sliceName.selectors.slectorName`
- But we also agreed to keep the slice name in the selector function name, i.e. `getUiVisualizationType`

I'm not 100% sure what to do, but I do think we should look for a more ergonomic approach where we don't have to destructure things. We've probably got two equally valid options:
1. We export the `selectors` object and use that directly (i.e. `selectors.selectorName`). However if we do this then probably we should stop using the slice name in the selector functions. For example, instead of `getCurrentId` we'd have `getId` and then if you need/want to differentiate/ make things explicit then you do `import selectors as currentSelectors from `@store/current-slice`...
2. Or doing what we're already doing in the state-slices, which is prefixing all reducers and selectors with the slice name `setUiVisualizationType`. But then we should probably export all of these ones individually, so they can be imported directly as well. Perhaps there is also a "object spread export syntax available" we could use (i.e. `export { ...uiSlice.selectors }` ?

src/types/analytics/index.d.ts
edoardo: Actually it seems I don't get the autocomplete working correctly for the components properties.
I'll just leave it as is.

src/store/current-slice.ts
```diff
@@ -0,0 +1,25 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { EventVisualization } from '@types'
+
+type CurrentState = Partial<EventVisualization> | null
```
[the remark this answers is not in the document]
edoardo: Fixed.

PR #31, opened 2025-08-25
feat: implement View menu in toolbar

edoardo:
Part of [DHIS2-19821](https://dhis2.atlassian.net/browse/DHIS2-19821)

### Description

Implements the View menu in the Menu bar of the Toolbar.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

Screenshot 2025-08-25 at 13 18 53

[DHIS2-19821]: https://dhis2.atlassian.net/browse/DHIS2-19821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/toolbar/menu-bar/view-menu.tsx
```diff
@@ -0,0 +1,86 @@
...
+export const ViewMenu: FC = () => {
+    const dispatch = useAppDispatch()
+
+    const isSidebarHidden = useAppSelector(getUiSidebarHidden)
+    const isLayoutPanelHidden = useAppSelector(getUiLayoutPanelHidden)
+    const isDetailsPanelOpen = useAppSelector(getUiDetailsPanelOpen)
+    const userSettingWidth = useAppSelector(getUiAccessoryPanelWidth)
+    const id = useAppSelector(getCurrentId)
+
+    const toggleLayoutPanelHidden = useCallback(() => {
```
edoardo, line 33: I guess in these cases we don't need to use `useCallback`, what do you think?

```suggestion
    const toggleLayoutPanelHidden = () => dispatch(toggleLayoutPanelHidden())
```

HendrikThePendric: This is a topic for heated debates. `dispatch` is a stable reference, so the `toggleLayoutPanelHidden` version that does `useCallback` and only has `dispatch` in its dependency array is also a stable reference. I think this is nice, because now if you pass it to child components, use it in a `useEffect` hook, or whatever, it will never cause unexpected rerenders.

On the other hand, others say it is an anti-pattern. And could also make things slower. So you should always use inline functions that get recreated at every render cycle (like your suggestion), unless that causes issues (either rerendering or performance).

I personally tend to do pretty much the opposite these days. I basically use `useCallback` pretty much everywhere. The only exception is when I write JSX and I attach event listeners to built-in element (i.e.`input`, `button`, etc.), because for these I know for a fact that it won't ever cause issues.

In my personal experience, the most common type of issue in React is that components rerender or effects run when you don't want/expect them to. So I am always trying to keep everything "as stable as possible". This means I always use `useMemo` and `useCallback`. It may be a bit controversial though.

src/constants/panels.ts
```diff
@@ -0,0 +1,8 @@
+export const ACCESSORY_PANEL_TAB_INPUT = 'INPUT'
+export const ACCESSORY_PANEL_TAB_PROGRAM = 'PROGRAM'
+export const ACCESSORY_PANEL_TAB_YOUR = 'YOUR'
+export const ACCESSORY_PANEL_TAB_TRACKED_ENTITY = 'TRACKED_ENTITY'
```
HendrikThePendric, line 4: Perhaps this should be a string literal type called `AccessoryPanelTab`?
```suggestion
export type AccessoryPanelTab = 'INPUT' | 'PROGRAM' | 'YOUR' | 'TRACKED_ENTITY'
```

HendrikThePendric (approved):
Looks great. All code looks good to me, and I have explained why I would NOT commit the suggestion you made.

src/components/toolbar/menu-bar/view-menu.tsx
edoardo: Ok, I saw you used `useCallback` in other parts of the app and I did the same, but then realised in my case there are no dependencies so I had a doubt.
I'm fine with keeping it and have the same approach everywhere.

src/constants/panels.ts
edoardo: Decided to remove these from this PR as they are not relevant.

HendrikThePendric (approved):

HendrikThePendric (approved):

PR #36, opened 2025-08-27
feat: add ScrollBox and useScrollBoxWidth hook [DHIS2-19816]

HendrikThePendric:
Implements [DHIS2-19816]](https://dhis2.atlassian.net/browse/DHIS2-19816]) (part 2)

### Description

This an advanced scroll-box component, which is mainly intended to host the visualization data-table. It comes with the following features:
1. The container fills up its parent and starts showing scrollbars in either direction when the content becomes wider/higher than the container
2. It renders its children into a provider so they can subscribe to its container-width using the `useScrollBarWidth` hook. This width is required for displaying the pagination correctly. It may also be helpful elsewhere.
3. The `width` returned by the hook takes the presence of a scrollbar into account, so it returns the container element's `clientWidth`, and it also observes resizes of the content element so width changes are also reported when the available width changes due to a vertical scrollbar appearing/disappearing due to content-changes.
4. The value of the width is debounced to avoid rer-renders during an animation, for example when one of the sidebars opens/closes/resizes.

#### Addition of a utility library

In this PR I also installed [usehooks-ts](https://github.com/juliencrn/usehooks-ts), which was discussed with and approved by both @edoardo and @jenniferarnesen during a call.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Cypress component tests added
- [x] Docs added N/A
- [x] Tester approved (@edoardo )

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
         "query-string": "^9.2.1",
         "react": "^18.0.0",
         "react-dom": "^18.0.0",
-        "react-redux": "^9.2.0"
+        "react-redux": "^9.2.0",
+        "usehooks-ts": "^3.1.1"
     },
     "resolutions": {
         "@dhis2/app-runtime": "^3.14.4"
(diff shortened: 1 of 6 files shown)


edoardo (approved):
Looks good, nice way of testing this.

PR #54, opened 2025-09-12
feat: add tooltip with content to layout chips

jenniferarnesen:
Implements [DHIS2-20105](https://dhis2.atlassian.net/browse/DHIS2-20105)

_This PR was started by @jenniferarnesen but mostly implemented by @HendrikThePendric_

### Description

This PR adds tooltips to the chips, that show the selected dimension options. Since these display names come from the metadata, and it turned out that many of them were not available, a lot of work to the metadata store is also done in this PR:
- Options from the visualisation follow-up request are now added to the metadata store differently (better)
- Metadata items are no longer populated with fields not present in the inputs
- We do no longer treat initial metadata as protected, because sometimes these initial items receive additional fields from response data and the "smart merge" feature actually handles this well
- The analytics response data is now added to the store via a dedicated `addAnalyticsResponseMetadata` method, which ensures it is processed correctly and the correct info is extracted
- We were already doing something similar with the initial metadata in `getInitialMetadata`, and the visualisation response data in `setVisualizationMetadata`. Some  improvement have been made here so that even more metadata is created/extracted
- Various tweaks have been made to the metadata input/item types, to better reflect the data we actually receive
- A fix to the `useMetadataItems` hook was implemented so that it no longer crashes when the `metadataItems` array changes size
- The "metadata store dev tools" have two new methods `getMetadataStoreItems` (accepts an array of IDs), `findMetadataStoreItem`, `filterMetadataStoreItems` (accept a token string and does a partial match on `id` or `name`)

To introduce the actual feature, a lot of code was ported from LL to produce the tooltip content based on the `conditions`, `dimension` and metadata. This code has been converted from JavaScript to TypeScript and a part of it had to be refactored. The challenge we faced was that chips are rendered when the visualisation response comes in, but some of the display names may rely on the analytics response data, which comes in later. To have get these items to update automatically, we need to use one of the "reactive hooks" (`useMetadataItem` / `useMetadataItems`). This was refactored as follows:
- The `getConditionsTexts` function in  LL did was split into multiple helpers, to both identify the way the text should be extracted, and to do the actual text extraction.
- After splitting things up we could differentiate between "computed texts" and "metadata-based texts"
- The helpers that returned "computed texts" were kept as-is, and the helpers that produced "metadata-based texts" were rewritten so they return an array of IDs instead of names
- The `getConditionsTexts` was then replaced by a `useConditionsTexts`, which internally use `useMetadataItems` to ensure the texts are updated automatically when as new metadata comes in.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A
- [x] Tester approved (@edoardo)

---

### ToDos

No to block this PR, but in new PRs, we should look into the following (I've checked the checkboxes after creating the  JIRA tickets):

- [x] Refactor the `MetadataStore` so it is more flexible in terms of the input it accepts. We should also pass the object key when processing the data because currently we are "preprocessing: the input to workaround current limitations [DHIS2-20470](https://dhis2.atlassian.net/browse/DHIS2-20470)
- [x] Refactor the`Axis` component and the `TooltipContent` so it can use the reactive `useMetadaItem` / `useMetadaItems` hooks. This should be possible for both components, but for Axis the suffix field should be computed upfront, since that needs a reference to other dimensions [DHIS2-20469](https://dhis2.atlassian.net/browse/DHIS2-20469)

---

### Known issues

- This relates to the second ToDo: in theory, if the dimension names are populated by the analytics data (and not the visualization data) the `Axis` could be showing an ID even thought the dimension metadata does resolve eventually. As far as I could tell this seems to be a theoretical issue.

---

### Screenshots

Screenshot 2025-11-18 at 10 22 30

[DHIS2-20105]: https://dhis2.atlassian.net/browse/DHIS2-20105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-20470]: https://dhis2.atlassian.net/browse/DHIS2-20470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout-panel/chip.tsx
+++ b/src/components/layout-panel/chip.tsx
@@ -81,7 +81,6 @@ export const Chip: React.FC<ChipProps> = ({ dimension, axisId }) => {
                 >
                     {({ ref, onMouseOver, onMouseOut }) => (
                         <span
-                            id={Math.random().toString(36)}
                             ref={ref}
                             onMouseOver={onMouseOver}
                             onMouseOut={onMouseOut}


src/components/app-wrapper/metadata-helpers/__tests__/visualization.spec.ts
```diff
@@ -0,0 +1,461 @@
...
+                {
+                    dimension: 'dimension1',
+                    dimensionType: 'DATA_ELEMENT',
+                    valueType: 'TEXT',
+                    optionSet: { id: 'optionSet1' },
+                } as DimensionRecord,
+            ],
+            rows: [],
+            filters: [],
+        } as unknown as SavedVisualization
```
edoardo: Probably not that important, but `SavedVisualization` requires `id`.
Can `NewVisualization`/`CurrentVisualization` be used here instead?

src/components/plugin-wrapper/plugin-wrapper.tsx
```diff
@@ -29,11 +29,11 @@ export const PluginWrapper: FC<PluginWrapperProps> = ({
...
     const [hasAnalyticsData, setHasAnalyticsData] = useState(false)
 
-    const onResponseReceived = useCallback(
-        (args) => {
+    const onResponseReceived = useCallback<OnAnalyticsResponseReceivedCb>(
+        (items, dimensions, headers) => {
             setHasAnalyticsData(true)
 
-            onResponsesReceivedCb?.(args)
+            onResponsesReceivedCb?.(items, dimensions, headers)
```
edoardo: As mentioned before, I think this should be more "generic" and accept different parameters because `headers` only applies to LL (to be confirmed).

src/components/app-wrapper/metadata-helpers/__tests__/visualization.spec.ts
HendrikThePendric: Since this is a test for metadata extraction from the visualization response, the `SavedVisualization` type is in fact the correct one. I tried if adding an `id` field would enable me to skip the `as unkown` cast, but this object is lacking more fields to be a fully qualified `SavedVisualization`. I also tried `const visualization: Partial<SavedVisualization> = { ... }`, which doesn't cause issues here, but does cause issues when the variable is used as a function parameter of type `SavedVisualization`. Considering that a) we're (re)using the `SavedVisualization` type for the function parameter (and I think this is good practice), and b) we do not need to populate all fieds to successfully test things, the current `as unknown as SavedVisualization` is the most progmatic solution.

src/components/plugin-wrapper/plugin-wrapper.tsx
HendrikThePendric: Yes true, as discussed, we'll see how we deal with this once we have the PT in place.

PR #60, opened 2025-09-18
feat: line-list scrolling and pagination [DHIS2-20133]

HendrikThePendric:
Implements [DHIS2-20133](https://dhis2.atlassian.net/browse/DHIS2-20133)

### Description

In this PR I have implemented/fixed the scrolling behaviour and have added the pagination component. I have also tranfered the styled from LL into the new components where needed, and have implemented the `fontSize` and `displayDensity` options.

I also ended up doing some work on `ScrollBox`:
- This is now internal the the `LineList`, since we do not have any use cases for it elsewhere
- It now has a border (and the table border was removed)
- I have tweaked the behaviour of this component, so that its behaviour is now more in line with what we want for the `LineList` (this required the tests to be updated too):
    - When the wrapped content is smaller than the scroll-box's container, then the scroll-box shrinks to the size of its content (before it would always occupy all of the containers width/height)
    - As the wrapped content becomes bigger, the scroll-box grows until it reaches the containers width/height and once the wrapped content exceeds the container's width/height a/multiple scrollbars will show

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A
- [x] Tester approved (@edoardo)

---

### ToDos

- I still have the `LegendSet` feature to implement and after that I will update the Cypress and Vitest test suite. I think it makes most sense to have an "intergration test" for the full LineList presentational component and its features

---

### Screenshots

Screenshot 2025-09-18 at 16 04 50

[DHIS2-20133]: https://dhis2.atlassian.net/browse/DHIS2-20133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/line-list/line-list.tsx
```diff
@@ -56,49 +58,30 @@ export const LineList: FC<LineListProps> = ({
...
-        isInDashboard,
-        '\npager: ',
-        pager,
-        '\n======'
+    const colSpan = useMemo(
+        () => String(Math.max(analyticsData.headers.length, 1)),
+        [analyticsData.headers.length]
     )
+
+    console.log('\nisInDashboard: ', isInDashboard)
```
edoardo, line 84: Cam this be removed?

edoardo (approved):
Looking good 👍

PR #61, opened 2025-09-19
feat: add landing page DHIS2-20123

edoardo:
Implements [DHIS2-20123](https://dhis2.atlassian.net/browse/DHIS2-20123)

### Description

When starting the app or clicking on File -> New, a landing page is shown.
It contains a list of hints on how to get started with a new visualization.
If statistic data about the visualizations is available, a list of the 6 most viewed event visualizations is shown.
Each item in the list is a link which acts as a shortcut to open the visualization.

When a saved visualization is opened, the most viewed statistics are updated.

---

### Quality checklist

- [ ] Dashboard tested N/A
- [ ] Cypress and/or Jest tests added/updated
- [ ] Docs added N/A
- [ ] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/XXX>) N/A
- [ ] Tester approved (@HendrikThePendric )

---

### Known issues

- when switching to the landing page via File -> New, the list of most viewed visualizations is not refetched

---

### Screenshots
Screenshot 2025-09-19 at 15 55 32

### DIFF
--- a/src/api/data-statistics-api.ts
+++ b/src/api/data-statistics-api.ts
@@ -1,0 +1,84 @@
+import { api } from '@api/api'
+import type { BaseQueryApiWithExtraArg } from '@api/custom-base-query'
+import { parseEngineError } from '@api/parse-engine-error'
+import type {
+    EventVisualizationType,
+    FavoriteStatistics,
+    SavedVisualization,
+} from '@types'
+
+type EventVisualizationRecord = Partial<SavedVisualization> &
+    Required<Pick<SavedVisualization, 'id' | 'type'>>
+type FavoriteRecord = FavoriteStatistics &
+    Required<Pick<FavoriteStatistics, 'id' | 'name'>>
+type MostViewedRecord = FavoriteRecord & {
+    type: EventVisualizationType
+}
+
+export const dataStatisticsApi = api.injectEndpoints({
+    endpoints: (builder) => ({
+        getMostViewed: builder.query<MostViewedRecord[], void>({
+            async queryFn(_args, apiArg: BaseQueryApiWithExtraArg) {
+                const { appCachedData, engine } = apiArg.extra
+
+                const username = appCachedData.currentUser.username
+
+                try {
+                    const mostViewedData: MostViewedRecord[] = []
+
+                    const favoritesResponse = await engine.query({
+                        favorites: {
+                            resource: 'dataStatistics/favorites',
+                            params: {
+                                eventType: 'EVENT_VISUALIZATION_VIEW',
+                                pageSize: 6,
+                                ...(username ? { username } : {}),
+                            },
+                        },
+                    })
+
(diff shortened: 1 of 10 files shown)


src/api/data-statistics-api.ts
```diff
@@ -0,0 +1,84 @@
...
+                            )?.type
+
+                            if (type) {
+                                mostViewedData.push({
+                                    id: favorite.id,
+                                    name: favorite.name,
+                                    type,
+                                })
+                            }
+                        }
```
HendrikThePendric, line 74: You are having to fly through a lot of TypeScript hoops here and ultimately this is because of a design error in app-runtime. The response type of `dataEngine.query` should have been `unknown`, so we are able to define the response type ourselves. The fact that you have to also wrap the query in a "nested object" and you also have to consume this nested structure further complicates matters....

I've had a go at simplifying the code somewhat by doing the following:
1. Since we have to "wrap the object" for `.query()` and the Web API also returns a "wrapped object", we are interested in this doubly nested property `result.eventVisualizations.eventVisualizations`. I've done the double unwrapping in the destructuring statement
2. Since the core issue is that the response has type `JsonMap` instead of `unknown`, I fix that by first overriding to unknown and then setting the correct response type.

I'm not completely sure that what I am doing in the end is that much an improvement, but I'll share a code suggestion anyway:

```suggestion
                        const {
                            eventVisualizations: { eventVisualizations },
                        } = (await engine.query({
                            eventVisualizations: {
                                resource: 'eventVisualizations',
                                params: {
                                    fields: 'id,type',
                                    filter: `id:in:[${favorites.map(
                                        ({ id }) => id
                                    )}]`,
                                },
                            },
                        })) as unknown as {
                            eventVisualizations: {
                                eventVisualizations: EventVisualizationRecord[]
                            }
                        }

                        for (const favorite of favorites) {
                            const type = eventVisualizations.find(
                                (vis) => vis.id === favorite.id
                            )?.type

                            if (type) {
                                mostViewedData.push({
                                    id: favorite.id,
                                    name: favorite.name,
                                    type,
                                })
                            }
                        }
```

HendrikThePendric (approved):
Great use case for a custom endpoint, and you also seemed to have found the prefect solution to the refetching issue with the `refetchOnMountOrArgChange` option. I believe there are other ways to "invalidate the query result" as well, but I think this option is probably the most straightforward, since the results need to update whenever a user visits/creates/edits visualizations, and this is pretty much 100% of the time.

I've left one suggestions to perhaps improve how to type the nested `eventVisualizations` object, but if you don't see it as an improvement, I'd also understand and I'd be fine keeping things as they are.....

src/api/data-statistics-api.ts
edoardo: Solved with #62 .

HendrikThePendric (approved):

PR #63, opened 2025-09-22
feat: enable sort and pagination in LL plugin [DHIS2-20124]

edoardo:
Implements [DHIS2-20124](https://dhis2.atlassian.net/browse/DHIS2-20124)

### Description

This PR adds the callbacks for handling sorting and pagination in the LL plugin.
These callbacks cannot rely on the app's Redux store as they  should also work when the plugin is used outside the app (ie. in Dashboard app).

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### ToDos

- [x] handle the sorting when used in the app context while editing and saving a visualization

---

### Screenshots

Sorting:
Screenshot 2025-09-22 at 14 49 42

Pagination:
Screenshot 2025-09-22 at 14 50 03

[DHIS2-20124]: https://dhis2.atlassian.net/browse/DHIS2-20124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HendrikThePendric (commented):
Nice to see everything come together. I have tested the Netlify PR and it seemed a few things weren't working as expected. I have spotted some things that were not in-scope for the current PR, but I also noticed a few things that look like they are. I'll just list all of them below:
1. [OUT-OF-SCOPE] The loader UI does not cover just the table, but the entire container. This needs to be addressed.
2. [FIXED-ELSEWHERE] I've seen some issues with the pagination not looking right. This turned out to be caused by the fact that my connection to the backend kept disconnecting (https://test.e2e.dhis2.org/anly-42). I keep seeing the offline indicator and then the pagination breaks. Actually, now that I think of it, I'm pretty sure this is fixed in #64, see [here](https://github.com/dhis2/event-visualizer-app/pull/64/files#diff-7c963f3d0c80c547c37458c2b6b0f22e855081588e0d1505d0b9e5894d24917cR82-R101). When showing the tooltip I was not using the "renderProps" way, so I was rendering a `span` around the `td` and this breaks the layout.
3. [MAYBE-IN-SCOPE]: Pagination works, but sorting doesn't. I keep seeing failed requests like below. It seems it doesn't matter what column I click, it always sorts on `pe`. I guess this could be a problem where we parse the headers. The header-cell simply passes the `header.name` to the `dimension` property of the `SortFn`.

I've also left some minor comments in the code.

Here's a reproduction path for the sorting-issue I have encountered:
- Use the Netlify link and use the `https://test.e2e.dhis2.org/anly-dev` backend
- Open this saved vis: https://pr-63.event-visualizer.netlify.dhis2.org/#/waPjzoJyIQ9
- Click on the sort-icon for the BMI column
- The sorting does't happen and in the network tab there is a 409

```
{
    "httpStatus": "Conflict",
    "httpStatusCode": 409,
    "status": "ERROR",
    "message": "Header param `pe` does not exist",
    "errorCode": "E7230"
}```

edoardo:
> 3. [MAYBE-IN-SCOPE]: Pagination works, but sorting doesn't. I keep seeing failed requests like below. It seems it doesn't matter what column I click, it always sorts on `pe`. I guess this could be a problem where we parse the headers. The header-cell simply passes the `header.name` to the `dimension` property of the `SortFn`.

> Here's a reproduction path for the sorting-issue I have encountered:
> 
> * Use the Netlify link and use the `https://test.e2e.dhis2.org/anly-dev` backend
> * Open this saved vis: https://pr-63.event-visualizer.netlify.dhis2.org/#/waPjzoJyIQ9
> * Click on the sort-icon for the BMI column
> * The sorting does't happen and in the network tab there is a 409
> 
> ```
> {
>     "httpStatus": "Conflict",
>     "httpStatusCode": 409,
>     "status": "ERROR",
>     "message": "Header param `pe` does not exist",
>     "errorCode": "E7230"
> }```
> ```

This was indeed a bug in this PRs code.
When applying the new sorting I was spreading the original visualization object, not the transformed one.
The hook that builds the analytics request needs the transformed visualization object in order to correctly compute the `headers` query parameter (for time dimensions in particular).

HendrikThePendric (approved):

PR #67, opened 2025-10-01
chore: add mock-app-wrapper utlity for Vitest and Cypress Component tests

HendrikThePendric:
Implements [DHIS2-20112](https://dhis2.atlassian.net/browse/DHIS2-20112)

### Description

This has become a bit more elaborate than what I originally intended to build. But I think it is also way more I have basically produced 3 variations of the same thing:
1. A `MockAppWrapper` component, which is mainly intended for Cypress Component tests
2. A `renderWithAppWrapper` function that can be used in Vitest component tests
3. A `renderHookWithAppWrapper` function that can be used in Vitest hook tests

All three share a lot of the same code and benefits:
- The component can be wrapped in the same contexts during a test as it is when in an app. This means all hooks are available without having to provide any mock. This includes the api-hooks, and the redux store hooks.
- All the initial metadata, state, and app-cached data is available, but you can also override/add metadata, provide a partial store, and override/add query data
- The Vitest helper functions also add a `store` property to the returned object, which means you can easily manipulate the read and update the store state during a test. The `MockAppWrapper` component does not have this direct benefit, but usually in a Cypress test you end up writing a small helper component anyway and in here you will have access to `useAppStore` hook, so with very little overhead you end up being able to read/update the store here too.

I've been able to greatly simplify a few existing tests with these new helpers.

In general I would like to see us leaning less heavily on E2E tests and writing more Unit and Intergration tests. I think these new helpers will be of great help for the latter.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Known issue

- The current implementation uses the custom-data-provider` to populate the app-cached-data. The custom-data-provider resolves a promise with the data and briefly has loading state, and the app-cached-data provider does not render its children until the data is resolved. The result of this is that the render and renderHook helpers are async and need to be awaited. I don't think this is problematic, but if it is we could implement a mocked version of the app-cached-data-provider which resolves its data synchronously.
- The `StoreToLocationSyncer` is not included in the mocked-app-wrapper. I don't see a need to do so

---

[DHIS2-20112]: https://dhis2.atlassian.net/browse/DHIS2-20112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test-utils/__tests__/app-wrapper.spec.tsx
+++ b/src/test-utils/__tests__/app-wrapper.spec.tsx
@@ -1,3 +1,4 @@
+import { FetchError } from '@dhis2/app-runtime'
 import { screen, waitFor } from '@testing-library/react'
 import { describe, it, expect } from 'vitest'
 import { renderWithAppWrapper, renderHookWithAppWrapper } from '../app-wrapper'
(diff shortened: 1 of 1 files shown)


HendrikThePendric:
I was surprised to see so many lines to be **removed** by this PR, because I have mainly been adding things. However, it turned out this is due to the fact that we had some stale snapshots of a large table in the codebase.

edoardo (dismissed):

edoardo (approved):

PR #81, opened 2025-10-21
feat: implement right sidebar with interpretations [DHIS2-20013]

edoardo:
Implements [DHIS2-20013](https://dhis2.atlassian.net/browse/DHIS2-20013)

### Description

This PR implements the interpretations and details panel and the interpretation modal.
Most components used are from `analytics` (See https://github.com/dhis2/analytics/pull/1791).

This PR also addresses the missing sorting handling in the dashboard plugin.

#### Note about the sorting handling

It was a bit tricky to make the sorting work in all scenarios without extra renders/fetches, but I think I found a decent solution.

There are 3 scenarios where we need the sorting to work, 2 and 3 are similar but 1 is different.

1. App scenario: the sorting needs to be saved in the Redux store in `currentVis` and this naturally would cause the LL plugin to refetch the analytics data. This is achieved via the `onDataSorted` callback in `app.tsx`
2. Interpretation modal scenario: here the interpretation component from analytics receives the `PluginWrapper` as plugin component to use in the modal, but it cannot know any specific features of the plugin, so there is no sorting callback to leverage (because the props passed to `PluginWrapper` are defined in `analytics`)
3. Dashboard plugin scenario: this is similar to the Interpretation modal scenario, although in `dashboard-plugin.tsx` it would be possible to implement a `onDataSorted` callback

Ideally the sorting implementation should live either in `PluginWrapper` or, better, in `LineListPlugin`.
The latter is best because other plugins (PT, charts) which also use `PluginWrapper` might not need the `onDataSorted` callback, or the implementation might be different for those vis types.

The single implementation is then in the `LineListPlugin` and it needs to use local state (due to scenarios 2 and 3)
The problem was how to avoid a double fetch in scenario 1, because with local state there are 2 updates, one when saving the new sorting in local state and one when the `onDataSorted` callback updates `currentVis` in the Redux store.
Both trigger an analytics fetch.

The solution is to check if `onDataSorted` is passed in the props and if so, assume that the consumer of the plugin components handles the sorting update and a  `visualization` prop change is triggered.

For scenarios 2 and 3, `onDataSorted` is not passed and the handling uses the local state which computes a new value for `visualization` which is catched by the same `useEffect` as in scenario 1.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A
- [x] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/1791>)

---

### Known issues

- [x] the table in the modal does not scroll and if the table is long enough, the Hide interpretation button is hidden

---

### Screenshots

Interpretations and details panel open:
Screenshot 2025-10-21 at 16 37 37

Interpretation modal open:
Screenshot 2025-10-21 at 16 37 44

[DHIS2-19822]: https://dhis2.atlassian.net/browse/DHIS2-19822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-20013]: https://dhis2.atlassian.net/browse/DHIS2-20013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/interpretations.cy.ts
+++ b/cypress/e2e/interpretations.cy.ts
@@ -1,0 +1,182 @@
+import {
+    typeTextarea,
+    clickMenubarInterpretationsButton,
+    clickMenubarViewButton,
+    expectInterpretationsButtonToBeEnabled,
+    expectInterpretationFormToBeVisible,
+    expectInterpretationThreadToBeVisible,
+    expectTableToBeVisible,
+    openVisByName,
+    deleteVisualization,
+    saveVisualizationAs,
+    expectVisTitleToEqual,
+} from '../helpers'
+
+const TEST_CANCEL_LABEL = 'Cancel'
+const TEST_POST_INTERPRETATION_LABEL = 'Post interpretation'
+const TEST_WRITE_INTERPRETATION_LABEL = 'Write an interpretation'
+const TEST_INTERPRETATION_TEXT = 'Test interpretation'
+
+describe('interpretations', () => {
+    beforeEach(() => {
+        cy.visit('/')
+    })
+
+    it('the interpretations button and item in View menu are disabled without a saved visualization', () => {
+        cy.getByDataTest('dhis2-analytics-toolbar')
+            .contains('Interpretations and details')
+            .should('be.disabled')
+
+        clickMenubarViewButton()
+
+        cy.getByDataTest('dhis2-uicore-hovermenulistitem')
+            .contains('Show interpretations and details')
+            .parent()
+            .should('have.class', 'disabled')
+    })
+
+    it('the interpretations and details panel can be toggled by clicking the button in the menu bar', () => {
+        openVisByName('Inpatient: Cases last quarter (case)')
+
+        expectInterpretationsButtonToBeEnabled()
+
+        cy.getByDataTest('details-panel').should('not.exist')
+
+        clickMenubarInterpretationsButton()
+
+        cy.getByDataTest('details-panel').should('be.visible')
+
+        cy.getByDataTest('details-panel').contains('Interpretations')
+        expectInterpretationFormToBeVisible()
+
+        clickMenubarInterpretationsButton()
+
+        cy.getByDataTest('details-panel').should('not.exist')
+    })
+
+    it('the interpretations and details panel can be toggled by clicking the option in the view menu', () => {
+        openVisByName('Inpatient: Cases last quarter (case)')
+
+        cy.getByDataTest('details-panel').should('not.exist')
+
+        clickMenubarViewButton()
+
+        cy.getByDataTest('dhis2-uicore-hovermenulistitem')
+            .contains('Show interpretations and details')
+            .should('be.visible')
+            .click()
+
+        cy.getByDataTest('details-panel').should('be.visible')
+
+        cy.getByDataTest('details-panel').contains('Interpretations')
+        expectInterpretationFormToBeVisible()
+
+        clickMenubarViewButton()
+
+        cy.getByDataTest('dhis2-uicore-hovermenulistitem')
+            .contains('Hide interpretations and details')
+            .should('be.visible')
+            .click()
+
+        cy.getByDataTest('details-panel').should('not.exist')
+    })
+
+    it('a new interpretation can be added, viewed and deleted', () => {
+        openVisByName('Inpatient: Cases last quarter (case)')
+
+        // Make a copy of the visualization
+        const visTitle = `INTERPRETATIONS TEST ${new Date().toLocaleString()}`
+        saveVisualizationAs(visTitle)
+
+        expectVisTitleToEqual(visTitle)
+        expectTableToBeVisible()
+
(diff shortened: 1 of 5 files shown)


src/components/app-wrapper/app-cached-data-query-provider.tsx
```diff
@@ -109,6 +110,9 @@ const providerDataTransformation = ({
             uiLocale: currentUser.settings?.keyUiLocale ?? 'en',
             displayProperty: currentUser.settings?.keyAnalysisDisplayProperty,
             displayNameProperty,
+            // preserve this original key used in the InterpretationModal component in analytics
+            keyAnalysisDisplayProperty:
+                currentUser.settings?.keyAnalysisDisplayProperty,
```
HendrikThePendric: Perhaps it would be nicer if we actually just address this in the InterpretationsProvider?

src/components/app-wrapper/interpretations-provider.tsx
HendrikThePendric: If we leave the analytics component unchanged, we should have a modified current user here.

src/components/app/app.tsx
HendrikThePendric: I'll leave a comment specifically related to the current PR as well, but I'm gonna start with a general one. 

I feel like having too much logic in this `App` component is not a good idea. In my mind this component should simply be a place where we gather all the components together. Since the App doesn't have a specific responsibility, having logic in here will make it hard follow, because it's going to be about a lot of different stuff. We should consider refactoring this.

The fact that this is a Redux based app should, in theory, put us in a perfect place to do so. We might want to create a few thunks for the callback that are currently being used by both the app and the interpretations modal. I do realise this will, for some components lead to slightly repetitive code, but as long as that is restricted to using the same selectors and thunks, I think it's a small price to pay to avoid chaos.

src/components/app/app.tsx
```diff
@@ -84,13 +94,35 @@ const EventVisualizer: FC = () => {
...
+                                    dispatch(
+                                        setNavigationState({
+                                            visualizationId: savedVis.id,
+                                            interpretationId: undefined,
+                                        })
+                                    )
+                                }}
+                                onResponsesReceived={onResponsesReceived}
+                            />
+                        )}
```
HendrikThePendric: I feel like this JSX plus all the selectors/callbacks/thunks it uses should become a dedicated component

HendrikThePendric (commented):
I know this is still in draft, but I left a review with some thoughts.

src/components/app-wrapper/app-cached-data-query-provider.tsx
edoardo: I think it makes sense.
I also don't like to have to add this "duplicated" key in `currentUser`.

src/types/analytics/about-ao-unit.ts
```diff
@@ -0,0 +1,9 @@
+import type { FC } from 'react'
+
+type AboutAOUnitProps = {
+    id: string
+    type: string
+    renderId?: number
```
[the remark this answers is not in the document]
edoardo: We agreed this is not in the scope of this PR.
It is relevant for the FileMenu actions PR where we do want to refresh the information shown in the about panel to  reflect changes in description and sharing settings.

We can also consider a refactor of the analytics component but that is a separate task.

src/components/app/app.tsx
edoardo: The `InterpretationModal` has now been extracted in a separate component in the app which renders the analytics modal component.
At the same time a provider has been added to handle `initialFocus` internally without needing a query string argument in the URL or something in the Redux state.

src/components/interpretation-modal/interpretation-modal.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const InterpretationModal: FC = () => {
+    const savedVis = useAppSelector(getSavedVis)
+    const { interpretationId, initialFocus } = useInterpretationModalState()
+    const { onCloseInterpretationModal } = useInterpretationModalTogglers()
+    const isVisualizationLoading = useAppSelector(getIsVisualizationLoading)
+
+    // No need to store metadata from the interpretaion modal analytics response
+    // The metadata is not accessed by the plugin anyway
+    const noop: (metadata: MetadataInput) => void = () => {}
```
HendrikThePendric: I see why we need to do this: `onResponseReceived` is a required prop in the `InterpretationModal` in analytics... So this makes sense. But perhaps it's better to simply declare the `noop` in the module scope so we don't run the risk to get an unnecessary unstable reference.

And perhaps clarify with a comment that this callback is a required prop.

edoardo: Thanks for the heads up.
Fixed.

HendrikThePendric (approved):

PR #83, opened 2025-10-31
refactor: add LayoutPanel component

edoardo:
### Description

This PR is a refactor of the layout component.
Now a `LayoutPanel` component is added, which renders the specific layout component depending on the visualization type.
This component also handles the show/hide feature where the whole `LayoutPanel` can be toggled via the View menu and/or the expanded canvas button.

The existing `LineListLayout` component is adjusted to use "group" divs which have the styles for the positioning and partitioning of the layout container space.
The existing `Axis` component is also simplified to simply render the chips and not have the logic for the positioning.

This grouping allows for more flexibility in the positioning and partitioning of the layout container.

The PR also adds an initial working version of the PT layout with Columns, Rows and Filters axis.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

LL layout:
Screenshot 2025-10-31 at 16 00 37

PT layout:
Screenshot 2025-10-31 at 16 03 08

### DIFF
--- a/src/components/layout-panel/layout-panel.tsx
+++ b/src/components/layout-panel/layout-panel.tsx
@@ -18,6 +18,7 @@ export const LayoutPanel: FC = () => {
         <div
             className={cx(classes.container, {
                 [classes.hidden]: !isLayoutPanelVisible,
+                [classes.lineList]: visualizationType === 'LINE_LIST',
             })}
         >
             <Axis axisId="columns" dimensionIds={columns} />
--- a/src/components/layout-panel/styles/layout-panel.module.css
+++ b/src/components/layout-panel/styles/layout-panel.module.css
@@ -9,6 +9,10 @@
     box-shadow: 0 1px 2px 0 rgb(0 0 0 / 0.03);
 }
 
+.container.lineList {
+    grid-template-areas: 'columns filters';
+}
+
 .container.hidden {
     display: none;
 }


HendrikThePendric (commented):
I have recently discovered the CSS Grid Layout and was so impressed with it that I now see it as a solution to every layout problem. So perhaps, it's better to evaluate for yourself instead of blindly following my suggestion below 😅 

I'm thinking that there are some advantages to using CSS grid layout for this:
- The PT Layout is 2 column, where the 1st column has 2 rows and the 2nd column just one. This can be done without any wrapper components when using CSS Grid. Having said that, it could be nice to still have wrapper components in the layout panel so that the Axis do not need to know anything about the Layout, they should just fill their container.
- The LL layout is just two columns so technically a flexbox layout should be fine. However, if we implement both layouts via CSS grid, then I'd think we wouldn't even need to have these separate LL and PR layouts, instead we'd:
    - Adjust the container CSS classname to a different grid definition
    - Conditionally render the "rows" `Axis`

edoardo:
> I have recently discovered the CSS Grid Layout and was so impressed with it that I now see it as a solution to every layout problem. So perhaps, it's better to evaluate for yourself instead of blindly following my suggestion below 😅
> 
> I'm thinking that there are some advantages to using CSS grid layout for this:
> 
> * The PT Layout is 2 column, where the 1st column has 2 rows and the 2nd column just one. This can be done without any wrapper components when using CSS Grid. Having said that, it could be nice to still have wrapper components in the layout panel so that the Axis do not need to know anything about the Layout, they should just fill their container.
> * The LL layout is just two columns so technically a flexbox layout should be fine. However, if we implement both layouts via CSS grid, then I'd think we wouldn't even need to have these separate LL and PR layouts, instead we'd:
>   
>   * Adjust the container CSS classname to a different grid definition
>   * Conditionally render the "rows" `Axis`

Cool!
I didn't use CSS grid before so I gave it a try and yes, it avoids quite a bit of stuff to obtain the (almost) same result.
I went for the areas approach thinking that it would be easier if we end up having to reposition them.
The only problem I see is that for LL which does not have Rows, the gap between Columns and Rows areas is still visible and noticeable where the Columns area touches the Filter one.
There are some solutions for this but I first wanted you to have a look at the changes.

src/components/layout-panel/styles/axis.module.css
```diff
@@ -1,19 +1,20 @@
...
+    grid-area: columns;
+}
+
+.axisContainer.rows {
+    grid-area: rows;
+}
+
+.axisContainer.filters {
+    grid-area: filters;
 }
```
HendrikThePendric: I mentioned earlier that I wasn't sure if we should have the `grid-area` definitions on the `Axis` component or if we should add wrapper elements to the `LayoutPanel` for a proper "separation of concern. I see you've added it to the `Axis` and I think this is a good call, since this is actually a sub-component, colocated with the `LayoutPanel`.

HendrikThePendric (commented):
Looks great @edoardo 🎉 

I first reviewed the code and then realised you also left a comment with a discussion point. I happened to have addressed that point in [one of my review-comments](https://github.com/dhis2/event-visualizer-app/pull/83#discussion_r2517383606). So I think this needs one small change and then we're good to merge it.

PR #105, opened 2025-12-01
feat: implement options for PT [DHIS2-20518]

edoardo:
Implements [DHIS2-20518](https://dhis2.atlassian.net/browse/DHIS2-20518)

### Description

Implement the remaining option components for PT.
Add configuration of Data and Style sections for PT.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### TODO

- [x] check if/how we need to handle disabled options

---

### Screenshots

Data section:
Screenshot 2025-12-03 at 16 28 26

Style section:
Screenshot 2025-12-02 at 15 44 36

[DHIS2-20518]: https://dhis2.atlassian.net/browse/DHIS2-20518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2025-12-03T15:26:17.163Z\n"
-"PO-Revision-Date: 2025-12-03T15:26:17.164Z\n"
+"POT-Creation-Date: 2025-12-04T16:22:54.198Z\n"
+"PO-Revision-Date: 2025-12-04T16:22:54.199Z\n"
 
 msgid "User organisation unit"
 msgstr "User organisation unit"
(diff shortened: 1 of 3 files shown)


HendrikThePendric (commented):
Sorry for not flagging this up in the previous PR. But I think we should make two general changes:
1. We should improve the naming conventions
2. We should extract everything into components, including the elements that just have a `className` and/or `data-test`

I think once we've decided on a clear naming scheme, it will be clear what components we extract. What is currently unclear is that the word "section" in the current implementation is ambiguous, it currently has the following meanings:
1. "tab content" 
2. "a grouped set fields"

I am open to changing things around in whichever way, as long as we end up with a distinct naming scheme. I'll just list some options we have available:
1. Instead of using "section" for "tab-content" we could simply switch to "tab-content". This would purely be a naming convention change.
2. Instead of using "section" for "a grouped set of fields" we could use "field-set". I like this option because it is also semantically correct: we could use a `fieldset` element and the "title" could be a `legend`. And then we could have a `OptionsFieldSet` component that accepts `children`, `title`, and potentially `helpText`, essentially the same as what [we have here](https://github.com/dhis2/analytics/blob/master/src/components/Options/VisualizationOptions.js#L44-L60). The fact that we were already using similar makes me think this is a viable option, but I should flag up that going this way could potentially cause some issues too, if we end up having to build new options UIs that are using a `fieldset` internally, since it is not recommended to nest `fieldset`s.

IMO, we could do 1, 2 or both, but we need to do at least one of them, because currently things are confusing.

I'm also considering that we could merge the current PR as-is, and then defer this work to a subsequent PR, because you could argue it's actually [DHIS2-20519](https://dhis2.atlassian.net/browse/DHIS2-20519).

Oh.... and we must also decide how to deal with the lack of backend support for that one field. Whichever option we choose there, we'll need a reference to JIRA ticket in this PR before we merge this.

HendrikThePendric (dismissed):
This looks perfect, great work.

PR #112, opened 2025-12-10
feat: implement Dimension modal component [DHIS2-20667]

edoardo:
Implements [DHIS2-20667](https://dhis2.atlassian.net/browse/DHIS2-20667)

### Description

This PR adds the DimensionModal component.
This renders the modal, modal title and action buttons.
The content for the modal is going to be implemented in several components depending of the dimension type.

The PR also has some fixes on types.

The Chip menu actions are now connected to the Redux store, so it's possible to move a dimension between the layout axes and remove a dimension from any axes.
The same reducer action is used for the split button in the dimension modal, which shows for dimensions that are not yet added to the layout (ie. in the main sidebar).

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### ToDos

- [x] check the `GenericMetadataItem` type I added to avoid type errors in `normalisation.spec.ts`.

---

### Screenshots

Examples of dimension modal opened:
Screenshot 2025-12-10 at 16 17 53
Screenshot 2025-12-10 at 16 17 47
Screenshot 2025-12-10 at 16 17 41

[DHIS2-20667]: https://dhis2.atlassian.net/browse/DHIS2-20667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2025-12-15T12:37:41.376Z\n"
-"PO-Revision-Date: 2025-12-15T12:37:41.377Z\n"
+"POT-Creation-Date: 2025-12-15T14:16:13.113Z\n"
+"PO-Revision-Date: 2025-12-15T14:16:13.113Z\n"
 
 msgid "User organisation unit"
 msgstr "User organisation unit"
@@ -282,6 +282,15 @@ msgstr "Normal"
 msgid "Compact"
 msgstr "Compact"
 
+msgid "Fix column headers to top of table"
+msgstr "Fix column headers to top of table"
+
+msgid "Fix row headers to left of table"
+msgstr "Fix row headers to left of table"
+
+msgid "There aren’t any row headers because Rows is empty."
+msgstr "There aren’t any row headers because Rows is empty."
+
 msgid "Font size"
 msgstr "Font size"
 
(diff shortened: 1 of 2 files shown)


src/components/app/app.tsx
```diff
@@ -85,6 +94,9 @@ const EventVisualizer: FC = () => {
                 >
                     Main sidebar
                 </div>
+                {activeDimensionModal && (
```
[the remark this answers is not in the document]
HendrikThePendric: Yeah I think this is OK for now. As I mentioned before, I do feel like what we're doing in `app.tsx` won't scale well. We keep adding small chunks of logic and JSX here with completely different scopes. I think that we will see two phases during the app's development:
1. It nice to have a single place where we can kind of see "how everything fits together"
2. The `app.jsx` file becomes too big and is doing too many different things and this is becoming troublesome. Once we get to this stage we should probably refactor the app component so that it is just a simple representational component that composes the full UI.

We are currently in phase 1 still, so this is fine. Also, if we did decide that we want to move to phase 2, we should do this in a dedicated PR.

src/types/layout.ts
```diff
@@ -0,0 +1,5 @@
+export type LayoutType = {
```
HendrikThePendric: It's good we have a type for this now, but why suffix it with `Type`, why not just `Layout`?

src/store/vis-ui-config-slice.ts
```diff
@@ -35,9 +33,9 @@ export const initialState: VisUiConfigState = {
     options: DEFAULT_OPTIONS,
     outputType: 'EVENT',
     layout: {
-        columns: [],
-        filters: [],
-        rows: [],
+        columns: EMPTY_STRING_ARRAY,
+        filters: EMPTY_STRING_ARRAY,
+        rows: EMPTY_STRING_ARRAY,
```
HendrikThePendric: Why is this preferable over `[]`?

edoardo: Changed back. We agreed is more readable and there isn't any reason to use the constants since this is an initial value.

PR #114, opened 2025-12-18
feat: implement modal content for OrgUnit dimensions [DHIS2-20681]

edoardo:
Implements [DHIS2-20681](https://dhis2.atlassian.net/browse/DHIS2-20681)

### Description

Implementation of the OrgUnit dimension modal content.
It uses the `OrgUnitDimension` component from `analytics`.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots
Screenshot 2025-12-18 at 14 04 59

[DHIS2-20681]: https://dhis2.atlassian.net/browse/DHIS2-20681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dimension-modal/orgunit-dimension-modal-content.tsx
+++ b/src/components/dimension-modal/orgunit-dimension-modal-content.tsx
@@ -1,4 +1,4 @@
-import { type FC } from 'react'
+import { useCallback, useMemo, type FC } from 'react'
 import { OrgUnitDimension, ouIdHelper } from '@dhis2/analytics'
 import {
     useAddMetadata,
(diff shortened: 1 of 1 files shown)


src/components/dimension-modal/orgunit-dimension-modal-content.tsx
```diff
@@ -0,0 +1,115 @@
...
+        if (metadata) {
+            acc.push({
+                id,
+                name: metadata?.name,
+                path: metadata?.path,
+            })
+        }
+
+        return acc
+    }, [] as OrgUnit[])
```
HendrikThePendric: I'm slightly worried about stability:
- `selectedIds` is probably stable, since it uses `useAppSelector`
- `selectedIdsMetadata` is probably stable too. Even if the array gets recreated I should return a stable reference from the metadata store if the array values remain constant
- `selected` should probably be wrapped in `useMemo`, since there is nothing preventing this array from being recreated on each component render, which is then passed to the `OrgUnitDimension`

🤔 I suppose it probably doesn't matter as long as the `dimension` prop doesn't change.

And also, now that I think about it, the same is also true for the `roots` and `onSelect` prop. 

I think it's probably all fine.

HendrikThePendric (approved):
Left a few non-critical comments

src/components/dimension-modal/orgunit-dimension-modal-content.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    return dimension ? (
+        <OrgUnitDimension
+            selected={selected}
+            roots={rootOrgUnits.map((rootOrgUnit) => rootOrgUnit.id)}
+            displayNameProp={currentUser.settings['displayNameProperty']}
+            onSelect={updateOrgUnitDimensionItems}
+        />
+    ) : null
```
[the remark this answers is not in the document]
edoardo: This was a leftover from when I started porting the components from LL.
Since we changed the approach here and let `DimensionModal` render the various content components we don't actually need `dimension`.
In LL the various content components render `DimensionModal` and `dimension` is needed there for setting the modal's title with the `dimension.name`.

Removed the ternary also in `period-dimension-modal-content`.

src/components/dimension-modal/orgunit-dimension-modal-content.tsx
edoardo: `selected` should only change if `visUiConfig.itemsByDimension` change for `dimension`, for example when changing/making a selection.
In that case the component re-renders and `selected` is recomputed.
Copilot suggests to write a memoized selector for `selectedIds`  because of the map used there.
I think the `selectedIds` can be actually passed from `DimensionModal` if this map is always needed.

PR #169, opened 2026-03-25
feat: custom value update button and modal [DHIS2-21124]

edoardo:
Completes [DHIS2-21124](https://dhis2.atlassian.net/browse/DHIS2-21124)

### Description

Implement the custom value button for PT and the modal for selecting the data element and the aggregation type to use.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

Empty state:
Screenshot 2026-03-31 at 16 37 40

Modal:
Screenshot 2026-03-27 at 14 36 32

Ready state (with tooltip):
Screenshot 2026-03-31 at 16 36 30

Screenshot 2026-03-31 at 16 38 42

[DHIS2-21124]: https://dhis2.atlassian.net/browse/DHIS2-21124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/store/thunks.ts
```diff
@@ -102,9 +102,10 @@ export const tUpdateCurrentVisFromVisUiConfig: AppThunk =
     () => (dispatch, getState) => {
         const { currentVis, visUiConfig } = getState()
 
-        const mergedVis = deepmerge(currentVis, visUiConfig.options) as
-            | NewVisualization
-            | SavedVisualization
+        const mergedVis = deepmerge(
+            currentVis as Record<string, unknown>,
+            visUiConfig.options as Record<string, unknown>
+        ) as unknown as NewVisualization | SavedVisualization
```
HendrikThePendric, line 108: This is a bit of a code smell I think. My hunch is that this is due to the fact that visUiConfig and currentVis have drifted in shape now. But considering the changes we are panning on doing, reg. the shape of the currentVis, I think we don't have to worry about this now. This code will be revised in the next few days anyway.

src/store/vis-ui-config-slice.ts
```diff
@@ -35,6 +43,8 @@ export interface VisUiConfigState {
     layout: Layout
     itemsByDimension: Record<string, string[]>
     conditionsByDimension: Record<string, ConditionsObject | undefined>
+    customValue?: CustomValueObject
+    lastActiveButton?: LastActiveButton
```
HendrikThePendric, line 47: It seems to me that the addition of the `customValue` field is completely justified. But I wonder if perhaps it would be possible to get rid of the `lastActiveButton` field and compute this in another way... Shouldn't we have `customValue` on the `currentVis` as well? And then the `lastUsedButton` is derived from the custom value there?

src/components/layout-panel/bottom-bar/action-buttons/use-action-button.ts
```diff
@@ -177,10 +191,17 @@ const getTrackedEntityInstanceTooltipContent = ({
     })
 }
 
-export const useActionButton = (buttonType: OutputType) => {
+export const useActionButton = (
+    buttonType: OutputType,
+    buttonVariant?: LastActiveButton
+) => {
```
HendrikThePendric, line 197: I suppose this indirectly relates to the other comments I left. The difference between `buttonType` and `buttonVariant` is not completely clear to me.

HendrikThePendric (commented):
The code here looks good. But some of the changes don't make sense to me by purely looking at the code changes and the PR description. I left a few comments in particular parts of the code, and these all seem to relate to the same underlying change... Not sure if I am flagging up a valid issue in the implementation, of if I am actually misunderstanding some of things you've done in this PR.

Up to you if you just want to talk about this via PR comments or have a call.

src/store/vis-ui-config-slice.ts
edoardo: Computing this based purely on `customValue` does not work because we don't clear `customValue` in Redux when we switch to a different button.

src/store/thunks.ts
edoardo: I think this was Claude fix for the TS problem, but I cannot remember why it started to happen...

src/components/layout-panel/bottom-bar/action-buttons/use-action-button.ts
edoardo: This was also Claude solution.
The problem is that for both Event and Custom value buttons, the `outputType` is `EVENT`, so technically they share the same analytics endpoint/request (except for the extra `value` argument).
I needed a way to distinguish the 2 buttons when `buttonType` (which determines `outputType`) is the same.
The custom button needs a different label, and also behaves differently, being a split button with 2 actions and having to open the modal.

HendrikThePendric (dismissed):
As agreed on Slack, there is some room for potential improvement here, but we'll defer that until we refactor the `currentVis` shape a bit.

HendrikThePendric (approved):

PR #183, opened 2026-04-09
feat: panel visibility dependencies [DHIS2-21277]

HendrikThePendric:
Implements [DHIS2-21277](https://dhis2.atlassian.net/browse/DHIS2-21277)

### Description

It was not completely clear in the specs how this should work exactly, but it was implemented as follows:
- The toolbar has two icon-buttons (this part is clearly specced):
    - toggle detail-sidebar visibility
    - toggle "expanded canvas" mode (AKA fullscreen mode) 
- The "view" dropdown menu also offers various ways to toggle panel visibility. This part was not 100% clear from the specs, because the EVER App has different panels than LL, so "keeping it the same" was not really possible:
    - Toggle layout
    - Toggle dimensions sidebar
    - Reset dimensions sidebar width (not relevant for this PR)
    - Toggle details
    
The interactions between the various buttons and menu-items was implemented as follows:
1. Going fullscreen -> hide all panels
2. Exiting fullscreen -> show previous panels
3. Using view menu when fullscreen -> exit fullscreen, show item that was selected, hide others
4. When each panel was hidden individually, this is equivalent to fullscreen. So when "exiting fullscreen" -> go back to the initial visibility state

I have added @cooper-joe as a reviewer for this PR becuase he needs to assess whether the current implementation is correct:
1. Do we have the correct buttons and menu-items?
2. Are the interactions between those items implemented correctly?

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

[DHIS2-21277]: https://dhis2.atlassian.net/browse/DHIS2-21277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cooper-joe:
> Do we have the correct buttons and menu-items?

Yes.

> Are the interactions between those items implemented correctly?

Yes, perfect. 👏 

The only thing that remains, which I remember you mentioned, is the inconsistent animation of the two sidebars. I'll make a note to review that and suggest something, but I don't think it needs to block this.

cooper-joe (dismissed):

src/store/ui-slice.ts
```diff
@@ -40,23 +80,26 @@ export const uiSlice = createSlice({
         },
         setUiDetailsPanelVisible: (state, action: PayloadAction<boolean>) => {
```
edoardo: This can actually be turned into `toggleUiDetailsPanelVisible` to keep it the same as the other toggles.
It seems like in the 2 places where this is used we actually do a toggle passing the negated `isDetailsPanelVisible`.

edoardo (dismissed):

src/store/ui-slice.ts
HendrikThePendric: Great suggestion. I implemented this change in [1d7db74](https://github.com/dhis2/event-visualizer-app/pull/183/commits/1d7db742e21263cb85e8e1ea0cf169ab27e6042b). Could you re-approve so this can get merged?

edoardo (approved):

PR #190, opened 2026-04-15
refactor: visualization and dimension transformations [DHIS2-21235]

HendrikThePendric:
Implements [DHIS2-21235](https://dhis2.atlassian.net/browse/DHIS2-21235)

### Description

Refactors how dimension metadata is extracted from saved visualizations and how fixed dimension names/IDs are produced. Establishes clear boundaries between wire-format (API) data and app-local data in the Redux store, with bidirectional translation at the boundary.

#### Data boundaries

```mermaid
graph TD
    API["eventVisualizations API"]

    subgraph wire["Wire-format — API dimension IDs"]
        savedVis["<b>savedVis slice</b><br/><i>SavedVisualization</i>"]
        currentVis["<b>currentVis slice</b><br/><i>CurrentVisualization</i>"]
    end

    subgraph boundary["Boundary: API — app-local"]
        forward_rename["toAppLocalDimensions<br/><code>ou → enrollmentOu</code>"]
        forward_compound["getCompoundDimensionId<br/><code>dim + context → stageId.dim / programId.dim</code>"]
        forward_fixed["getStageFixedDimensions<br/>getEnrollmentFixedDimensions<br/>getTrackedEntityTypeFixedDimensions"]
    end

    subgraph appLocal["App-local — compound IDs, canonical names"]
        visUiConfig["<b>visUiConfig slice</b><br/><i>layout, items, conditions, repetitions</i>"]
        metadataStore["<b>MetadataStore</b><br/><i>keyed by compound IDs</i>"]
        sidebar["<b>Sidebar cards</b><br/><i>uses same fixed dimension helpers</i>"]
    end

    API -->|"normalize (legacy)"| savedVis
    savedVis -->|"toCurrentVis()"| currentVis
    currentVis -->|"getVisualizationUiConfig()"| forward_rename
    savedVis -->|"extractMetadataFromVisualization()"| forward_rename
    forward_rename --> forward_compound --> visUiConfig
    forward_rename --> forward_fixed --> metadataStore
    forward_fixed --> sidebar
    visUiConfig -->|"tUpdateCurrentVisFromVisUiConfig()"| currentVis
```

**Wire-format** (`savedVis`, `currentVis`): Use API dimension IDs (e.g. `ou` for enrollment org unit) with `program`/`programStage` context on each dimension record. These map directly to what the eventVisualizations API sends and expects.

[...]

PR #194, opened 2026-04-22
fix: fix analytics requests

edoardo:
Implements [DHIS2-XXXX](https://dhis2.atlassian.net/browse/DHIS2-XXXX)

### Description

The way the analytics request is constructed is different in EVER compared to LL.
The PR addresses the differences and new things according to the spreadsheet.

The main difference is that all dimensions with no conditions are not added as `dimension` parameters or "standalone" parameters. For those situations only `headers` is passed in the request.
For dimensions with conditions applied, `dimension` parameters are added for all dimensions (before some dimensions where passed as "standalone" parameters).

This PR has fixes also for the aggregate (PT) cases.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A
- [x] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/1815>)

---

### Screenshots

Screenshot 2026-04-20 at 13 49 12

### DIFF
--- a/src/components/app-wrapper/metadata-provider/analytics-data.ts
+++ b/src/components/app-wrapper/metadata-provider/analytics-data.ts
@@ -74,8 +74,13 @@ const updateNamesFromHeaders = (
 
 export const extractMetadataFromAnalyticsResponse = (
     items: AnalyticsResponseMetadataItems,
-    headers: Array<LineListAnalyticsDataHeader>
+    headers?: Array<LineListAnalyticsDataHeader>
 ): MetadataInput => {
     const metdataFromItems = extractItemsMetadata(items)
-    return updateNamesFromHeaders(headers, metdataFromItems)
+
+    if (headers?.length) {
+        return updateNamesFromHeaders(headers, metdataFromItems)
+    }
+
+    return metdataFromItems
 }
--- a/src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
+++ b/src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
@@ -28,6 +28,12 @@ const metadata = {
         programType: 'WITH_REGISTRATION',
         trackedEntityType: { id: 'tei1' },
     },
+    p3: {
+        id: 'p3',
+        name: 'Program3',
+        programType: 'WITH_REGISTRATION',
+        trackedEntityType: { id: 'tei2' },
+    },
     tei1: {
         id: 'tei1',
         name: 'Person',
@@ -76,6 +82,8 @@ const metadata = {
         dimensionType: 'DATA_ELEMENT',
         optionSet: 'OptionSet1',
         valueType: 'TEXT',
+        programId: 'p2',
+        trackedEntityTypeId: 'tei1',
     },
     'p2.p2s1.d2': {
         id: 'p2.p2s1.d2',
@@ -94,6 +102,15 @@ const metadata = {
         name: 'Program indicator',
         dimensionType: 'PROGRAM_INDICATOR',
         valueType: 'TEXT',
+    },
+    'p3.p3s1.d1': {
+        id: 'p3.p3s1.d1',
+        name: 'Dimension1',
+        dimensionType: 'DATA_ELEMENT',
+        optionSet: 'OptionSet1',
+        valueType: 'TEXT',
+        programId: 'p3',
+        trackedEntityTypeId: 'tei2',
     },
     'tei1.created': {
         id: 'tei1.created',
@@ -896,6 +913,34 @@ describe('useActionButton for Tracked entity instance button', () => {
         expect(output.dataSourceMetadata).toEqual(metadata.p2)
         expect(output.tooltipConfig).toEqual({
             content: 'Not valid with multiple programs',
+        })
+    })
+
+    it.only('returns correct result for: PT, multiple TET', async () => {
+        const { result } = await renderHookWithAppWrapper(
+            () => useActionButton('TRACKED_ENTITY_INSTANCE'),
+            createStoreWithPreloadedState({
+                dimensionSelection: {
+                    dataSourceId: metadata.p2.id,
+                },
+                visUiConfig: {
+                    layout: {
+                        columns: [
+                            metadata['p2.p2s1.d1'].id,
+                            metadata['p3.p3s1.d1'].id,
+                        ],
+                    },
+                    visualizationType: 'PIVOT_TABLE',
+                },
+            })
+        )
+
+        const output = result.current
+
+        expect(output.action).toEqual('create')
+        expect(output.dataSourceMetadata).toEqual(metadata.p2)
+        expect(output.tooltipConfig).toEqual({
+            content: 'Not valid with multiple tracked entity types',
         })
     })
 
--- a/src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
+++ b/src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
@@ -917,7 +917,7 @@ describe('useActionButton for Tracked entity instance button', () => {
         })
     })
 
-    it.only('returns correct result for: PT, multiple TET', async () => {
+    it('returns correct result for: PT, multiple TET', async () => {
         const { result } = await renderHookWithAppWrapper(
             () => useActionButton('TRACKED_ENTITY_INSTANCE'),
             createStoreWithPreloadedState({
--- a/src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
+++ b/src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
@@ -79,9 +79,8 @@ const metadata = {
     'p2.p2s1.d1': {
         id: 'p2.p2s1.d1',
         name: 'Dimension1',
-        dimensionType: 'DATA_ELEMENT',
-        optionSet: 'OptionSet1',
-        valueType: 'TEXT',
+        dimensionType: 'ORGANISATION_UNIT',
+        valueType: 'ORGANISATION_UNIT',
         programId: 'p2',
         trackedEntityTypeId: 'tei1',
     },
@@ -176,7 +175,6 @@ describe('useActionButton for Event button', () => {
         const output = result.current
 
         expect(output.action).toEqual('switch')
-        expect(output.dataSourceMetadata).toEqual(metadata.p1)
         expect(output.tooltipConfig).toEqual(undefined)
     })
 
@@ -204,7 +202,6 @@ describe('useActionButton for Event button', () => {
         const output = result.current
 
         expect(output.action).toEqual('update')
-        expect(output.dataSourceMetadata).toEqual(metadata.p1)
         expect(output.tooltipConfig).toEqual(undefined)
     })
 
@@ -224,7 +221,6 @@ describe('useActionButton for Event button', () => {
         const output = result.current
 
         expect(output.action).toEqual('create')
-        expect(output.dataSourceMetadata).toEqual(metadata.p1)
         expect(output.tooltipConfig).toEqual({
             content:
                 'Nothing selected. Add items to the layout to get started.',
@@ -254,7 +250,6 @@ describe('useActionButton for Event button', () => {
         const output = result.current
 
         expect(output.action).toEqual('create')
-        expect(output.dataSourceMetadata).toEqual(metadata.p1)
         expect(output.tooltipConfig).toEqual({
             content: 'Not valid with multiple program stages',
         })
@@ -282,7 +277,6 @@ describe('useActionButton for Event button', () => {
         const output = result.current
 
         expect(output.action).toEqual('create')
-        expect(output.dataSourceMetadata).toEqual(metadata.p1)
         expect(output.tooltipConfig).toEqual({
             content: 'Not valid with multiple program stages',
         })
@@ -307,7 +301,6 @@ describe('useActionButton for Event button', () => {
         const output = result.current
 
         expect(output.action).toEqual('create')
-        expect(output.dataSourceMetadata).toEqual(metadata.p2)
         expect(output.tooltipConfig).toEqual({
             content: 'Not valid without a program',
         })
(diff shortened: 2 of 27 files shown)


src/components/app-wrapper/metadata-provider/analytics-data.ts
```diff
@@ -74,8 +74,13 @@ const updateNamesFromHeaders = (
 
 export const extractMetadataFromAnalyticsResponse = (
     items: AnalyticsResponseMetadataItems,
-    headers: Array<LineListAnalyticsDataHeader>
+    headers?: Array<LineListAnalyticsDataHeader>
 ): MetadataInput => {
     const metdataFromItems = extractItemsMetadata(items)
-    return updateNamesFromHeaders(headers, metdataFromItems)
+
+    if (headers?.length) {
```
edoardo, line 81: @HendrikThePendric this is because in PT we don't pass `headers` in the analytics request.

src/components/line-list/use-transformed-line-list-data.ts
```diff
@@ -66,7 +66,7 @@ const getHeaderDimensionId = (
     })
     const idMatch =
         Object.keys(headersMap).find(
-            (key) => headersMap[key] === dimensionId
+            (key) => headersMap[key] === header.name
```
edoardo, line 70: I'm not totally sure about this one.
`header.name` sometimes has a prefix, so this line wouldn't match...

src/components/line-list/use-transformed-line-list-data.ts
```diff
@@ -156,7 +157,7 @@ const getFormattedCellValue = ({
                 headersMap.enrollmentDate,
                 headersMap.incidentDate,
                 headersMap.scheduledDate,
-            ].includes(dimensionId)
+            ].includes(header.name)
```
edoardo, line 161: I think we need to review this whole header mapping.

src/components/main-sidebar/dimension-card/card-metadata.tsx
```diff
@@ -24,7 +24,6 @@ const getFixedDimensions = (): DimensionMetadataItem[] => [
         dimensionId: 'lastUpdatedBy',
         name: i18n.t('Last updated by'),
         dimensionType: 'USER',
-        valueType: 'USERNAME',
```
edoardo, line 27: This is so that the dimension modal does not show the condition filtering but a message saying all items are used.

src/components/plugin-wrapper/hooks/query-tools-line-list.ts
```diff
@@ -103,9 +147,16 @@ export const getAdaptedVisualization = (
 
     return {
         adaptedVisualization: {
-            columns: adaptedColumns,
-            rows: adaptedRows,
-            filters: adaptedFilters,
+            // only pass dimensions with conditions
+            columns: adaptedColumns.filter(
+                ({ filter, items }) => filter || items?.length
```
edoardo: We might need to change this.
For the dimensions listed in the "Other" card, not having an empty dimension in the analytics request causes errors.
Now it's too late to fix the API for 43 so we need to handle this here somehow.
I'm pretty sure while testing I found some issues with passing dimension parameters without a condition, but I cannot reproduce now, so we could try to simply remove this filter and test.
If it doesn't work with some dimensions we'll see it.

src/components/plugin-wrapper/hooks/query-tools-pivot-table.ts
```diff
@@ -0,0 +1,113 @@
...
+            const dimension = convertToScreamingSnakeCase(dimensionId)
+
+            // remove program/programStage for these dimensions
+            adaptedDimensions.push({
+                ...dimensionObj,
+                dimension,
+                program: undefined,
+                programStage: undefined,
+            })
+        } else if (dimensionId === 'ou') {
```
edoardo, line 57: This part (and the missing headers) is pretty much where the whole function differs from the one for LL.

src/components/plugin-wrapper/hooks/use-pivot-table-analytics-data.ts
```diff
@@ -126,7 +147,7 @@ const usePivotTableAnalyticsData = (): UseAnalyticsDataResult => {
                 })
 
                 // TODO: enable this once the metadata PR is merged
```
edoardo: I guess this can be removed. I understand it refers to your metadata PR.

src/components/layout-panel/bottom-bar/action-buttons/use-action-button.ts
```diff
@@ -307,23 +314,15 @@ export const useActionButton = (
...
-        if (isDataSourceProgramWithRegistration(dataSourceMetadata)) {
-            const tetId = dataSourceMetadata.trackedEntityType.id
-
-            return isDimensionInLayout(layout, `${tetId}.ou`)
-        }
+        const tetId = getTrackedEntityTypeIdFromDataSource(dataSourceMetadata)
 
-        return false
+        return tetId ? isDimensionInLayout(layout, `${tetId}.ou`) : false
     }, [dataSourceMetadata, layout])
```
HendrikThePendric: I think this relates to what we discussed today. We shouldn't be referring to the data source for the TET ID, but extract it from the layout.

As you know this is also a fundamental issue in the base branch, so I don't know what's best:
1. I solve it in the base branch and you solve it here
2. We add TODOs and solve it later on

src/components/line-list/use-transformed-line-list-data.ts
HendrikThePendric: I guess we can do `header.name.endsWith(headersMap[key])`?

src/components/line-list/use-transformed-line-list-data.ts
```diff
@@ -130,17 +130,18 @@ const DATE_VALUE_TYPES: ValueType[] = ['DATE', 'DATETIME']
...
     visualization,
 }: {
     value: string
-    dimensionId: string
     header: LineListAnalyticsDataHeader
     visualization: CurrentVisualization
 }) => {
+    // header.name might be prefixed with programStage.id
+    const dimensionId = header.name?.split('.').at(-1)
+
```
HendrikThePendric, line 143: I have a helper for this `extractPlainDimensionId`. But I've noticed we don't use this helper consistently yet.

src/components/line-list/use-transformed-line-list-data.ts
HendrikThePendric: Fair enough....

In general I think that before merging our open PRs we create some JIRA tickets or sth.

src/components/plugin-wrapper/hooks/query-tools-line-list.ts
```diff
@@ -7,46 +7,81 @@ import type {
     OutputType,
 } from '@types'
 import { getRequestOptions } from './query-tools-common'
-import type { ParameterRecord } from './query-tools-common'
+
+const convertToScreamingSnakeCase = (input: string) =>
+    input.replaceAll(/[A-Z]/g, (c) => `_${c}`).toUpperCase()
```
HendrikThePendric, line 13: I think this is probably OK, but please confirm that we never have any leading chars in uppercase. Because `TestStr` -> `_TEST_STR`

src/components/plugin-wrapper/hooks/query-tools-line-list.ts
```diff
@@ -7,46 +7,81 @@ import type {
...
 
 const adaptDimensions = (
     dimensions: DimensionArray,
-    parameters: ParameterRecord,
     outputType: OutputType
 ) => {
     const adaptedDimensions: DimensionArray = []
 
+    // This function should accommodate the various dimension cases and exceptions detailed in the Analytics apis spreadsheet
```
HendrikThePendric: Remove

src/components/plugin-wrapper/hooks/query-tools-line-list.ts
```diff
@@ -7,46 +7,81 @@ import type {
...
+                adaptedDimensions.push({
+                    ...dimensionObj,
+                    dimension,
+                    program: undefined,
+                    programStage: undefined,
+                })
+            }
+        } else if (dimensionId === 'enrollmentOu') {
+            // enrollmentOu must be passed as ou for ENROLLMENT
+            // program prefix must be removed for EVENT/ENROLLMENT
```
HendrikThePendric: Not 100% sure, but I think this is already handled upstream (I have code for `ou` <--> `enrollmentOu` in the base branch.

Once in the CurrentVis, the dimension should already have the correct ID.

src/components/plugin-wrapper/hooks/query-tools-pivot-table.ts
HendrikThePendric: Yeah I got a bit confused because these two files looked so similar, and started writing a comment about code duplication. I think we should extract some shared functions here.

And possibly, the fact that I should already be taking care of the `ou` <--> `enrollmentOu` renaming when we update the CurrentVis from the VisUiConfig, means we can maybe even have one shared function?

src/components/plugin-wrapper/hooks/use-pivot-table-analytics-data.ts
HendrikThePendric: Correct!

src/components/plugin-wrapper/pivot-table-plugin.tsx
```diff
@@ -1,10 +1,54 @@
...
+            let dimensionId: string
+
+            if (dimensionObj.dimension === 'ou') {
+                dimensionId = dimensionObj.programStage?.id
+                    ? dimensionObj.dimension
+                    : 'enrollmentou'
+            } else {
+                dimensionId =
+                    headersMap[dimensionObj.dimension] ?? dimensionObj.dimension
+            }
```
HendrikThePendric, line 32: This comment is similar to what I wrote before... I _should_ be taking care of this conversion already.

src/modules/visualization.ts
```diff
@@ -58,22 +59,27 @@ export const headersMap: Record<DimensionId, string> = {
...
+
     const map = Object.assign({}, headersMap)
 
     if (showHierarchy) {
         map['ou'] = 'ounamehierarchy'
     }
 
+    if (outputType === 'ENROLLMENT') {
+        map['enrollmentOu'] = 'ouname'
+    }
```
HendrikThePendric, line 81: Same as before

HendrikThePendric (commented):
I've left 11 comments. Quite a few are related to `ou` <--> `enrollmentOu` conversion, which I believe should be handled in the base branch already, so if this is not being done correctly, we should probably address it there rather than here. We should do it when we convert from `visUiConfig` -> `currentVis` and when we load a saved vis, because then we reduce the "surface area" to a minimum. In theory it should be once and in this PR we see it done a bunch of times.

We should also discuss a general plan forward. There clearly still is work to be done, i.e. on how we deal with header names etc. We should discuss what's in-scope and what's out-of-scope and we should create JIRA tickets for the out-of-scope work.

For this PR and #190 I am thinking that we may want to actually extend the scope a bit so that it includes:
1. Reading TET ID via org. unit dimensions instead of the selected data source
2. Header name handling (maybe)

src/components/plugin-wrapper/hooks/use-pivot-table-analytics-data.ts
edoardo: Removed the comment.

src/components/plugin-wrapper/hooks/query-tools-line-list.ts
edoardo: Yes, that shouldn't be the case.

src/components/line-list/use-transformed-line-list-data.ts
edoardo: Replaced with the helper function.

src/modules/visualization.ts
HendrikThePendric: Sorry I misread this one. This is valid.

src/components/plugin-wrapper/hooks/query-tools-pivot-table.ts
HendrikThePendric: As discussed this `ou` -> `ENROLLMENT_OU` is for the infamous L65 on the Google Sheet. This dimension renaming is indeed needed.

My comment about code re-use remains potentially valid though, so I am not resolving my comment yet.

src/components/plugin-wrapper/pivot-table-plugin.tsx
HendrikThePendric: OK I think this is L65 of the Google Sheet, so an invalid comment.

src/components/plugin-wrapper/hooks/query-tools-line-list.ts
edoardo: Removed the dimension override as this case should not happen.
`ou` would be correctly passed already.

src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
```diff
@@ -829,6 +917,35 @@ describe('useActionButton for Tracked entity instance button', () => {
         })
     })
 
+    it.only('returns correct result for: PT, multiple TET', async () => {
```
HendrikThePendric: The `.only` needs to be removed, and possibly this means some other tests will need to be fixed.

src/components/layout-panel/bottom-bar/action-buttons/enrollment-button.tsx
```diff
@@ -8,20 +9,21 @@ import { useActionButton } from './use-action-button'
...
     const visualizationType = useAppSelector(getVisUiConfigVisualizationType)
-
-    const { action, dataSourceMetadata, tooltipConfig } =
-        useActionButton('ENROLLMENT')
+    const { action, tooltipConfig } = useActionButton('ENROLLMENT')
+    const programIds = useProgramIds()
+    const programMetadata = useProgramMetadataItem(programIds[0])
 
     const enrollmentLabel = useMemo(() => {
+        console.log('pr md', programMetadata)
```
HendrikThePendric: Remove

HendrikThePendric (changes requested):
I spotted a critical issue in one of the test files: the `.only` needs to be removed so all tests are executed. But other than that this looks perfect to me.

src/components/layout-panel/bottom-bar/action-buttons/__tests__/use-action-button.spec.tsx
edoardo: Yep, forgot about it. Removed now.

HendrikThePendric (approved):

PR #203, opened 2026-05-05
chore: read tet id via org unit dimension

HendrikThePendric:
Implements: N/A (it's a fixup)

### Description

Resolves the `trackedEntityType` ID for updating the `currentVis` from the `visUiConfig` by walking the layout's organisation unit dimension instead of consulting the data source / scanning layout dims for a TEA's `trackedEntityTypeId`. 

Also fixes several related issues that surfaced while doing so:
- In the sidebar we were using `ou` instead of the "app-local" `enrollmentOu` for "Registration org. unit" (in the TET card for tracker-programs and TET data sources). This was fixed, we're using `enrollmentOu` now.
- This change then required updating the logic in `toApiDimensionId` so this was fixed as well. As discussed this function was also renamed to `toEventVisualizationDimensionId` for clarity.
- We also needed some updates to the inverse dimension ID handling (when we extract dimensions from the visualization), so now `TEI_REGISTRATION_DIMENSION_IDS` is updated from `['ou', 'created']` to `['enrollmentOu', 'created']` so `getCompoundDimensionId` reconstructs `${tetId}.enrollmentOu` on saved-vis load. `toAppLocalDimensions` was also simplified to handle the unqualified `ou` form (TEI registration OU on load).
- Fixed a hard-coded `${tetId}.ou` reference in `use-action-button.ts` that would have silently broken otherwise.
- I also spotted a function that was renamed and exported in `cards-tracked-entity-type/card-tracked-entity-type.tsx`. It was also being tested in the wrong place. 
- The Claude instructions have been compacted because I got a warning about the file exceeding the max. recommended size. I also made some specific changes:
    - "Org unit scopes" were updated for the new `tetId.enrollmentOu` form. 
    - Translation table extended with a TEI registration row gated on `programId` presence.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### DIFF
--- a/src/components/main-sidebar/cards-program-with-registration/card-tracked-entity-type.tsx
+++ b/src/components/main-sidebar/cards-program-with-registration/card-tracked-entity-type.tsx
@@ -12,13 +12,11 @@ import {
 } from '@components/main-sidebar/use-selected-dimension-count'
 import i18n from '@dhis2/d2-i18n'
 import { useCurrentUser } from '@hooks'
-import type {
-    DataSourceProgramWithRegistration,
-    DimensionMetadataItem,
-} from '@types'
+import { getTrackedEntityTypeFixedDimensions } from '@modules/dimension'
+import type { DataSourceProgramWithRegistration } from '@types'
 import { useCallback, useMemo, type FC } from 'react'
 
-export const transformProgramAttributes = (
+const transformProgramAttributes = (
     data: unknown,
     trackedEntityTypeId: string
 ): ReturnType<Transformer> => {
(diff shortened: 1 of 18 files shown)


src/components/main-sidebar/cards-program-with-registration/card-tracked-entity-type.tsx
```diff
@@ -62,8 +48,11 @@ export const CardTrackedEntityType: FC<CardTrackedEntityTypeProps> = ({
             program.trackedEntityType.name,
     })
     const fixedDimensions = useMemo(
-        () => getFixedDimensions(program),
-        [program]
+        () =>
+            getTrackedEntityTypeFixedDimensions(
+                program.trackedEntityType
+            ).filter((dimension) => dimension.dimensionId === 'enrollmentOu'),
```
edoardo, line 54: Does this mean that the only fixed dimension for TET is `enrollmentOu`?

edoardo (approved):

src/components/main-sidebar/cards-program-with-registration/card-tracked-entity-type.tsx
HendrikThePendric: Note that there are 2 TET cards:
1. The TET card we show when the data source is a tracker program. Here the only fixed dimension is `enrollemntOu`. In the design spec we also have "Registration period", but Jan asked me to drop that one, I forget why this was exactly.
2. The TET card we show then choosing the "analyse without a program" / TET data source. Here we do show both the "Registration org. unit" as well as the "Registration period".

So the answer to your question is: _it depends which TET_

PR #209, opened 2026-05-07
feat: chip suffixes and tooltip content [DHIS2-21466]

HendrikThePendric:
Implements [DHIS2-21466](https://dhis2.atlassian.net/browse/DHIS2-21466)

### Description

Implements chip suffixes as described in the [design specs](https://docs.google.com/document/d/1v0CRmuV-VBXc_-Xc-JP8oj3pMbJIVInLpf21KURXAzk/edit?tab=t.0#heading=h.ls0pg0sv6oct).

It also simplifies the logic for constructing the tooltip texts

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

Screenshot 2026-05-07 at 16 06 34

[DHIS2-21466]: https://dhis2.atlassian.net/browse/DHIS2-21466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/plugin-wrapper/hooks/use-line-list-analytics-data.ts
edoardo, line 268: It seems we have a problem here with some dimensions where the id does not match in the find.
ie. `stage.ou` vs `stage.ouname`.

Image
Image

Whatever goes wrong causes the header label to be wrong:

Image

edoardo (changes requested):
Overall it looks good, but when opening saved visualizations I found several where some column header labels are broken.

I think the part of code I commented on is involved.

src/components/plugin-wrapper/hooks/use-line-list-analytics-data.ts
edoardo: Actually, the problem seems to be before that, where `labels` is computed.
You can see here there is missing metadata for some dimensions like `stageId.eventStatus` and `stageId.ou`.
These seem to match the ones I noticed have broken column labels:
Screenshot 2026-05-08 at 11 52 57

PR #210, opened 2026-05-07
refactor: rename and reimplement various dimension ID helpers [DHIS2-21468]

HendrikThePendric:
Implements [DHIS2-21468](https://dhis2.atlassian.net/browse/DHIS2-21468)

### Description

While looking into this it became apparent that consolidation was not really possible, because we have distinct contexts:
- App code where we can only look things up via the metadata store because the `outputType` is unknown when the code is executed
- Plugin code where the `outputType` is known but the metadata store is not present

So while it was not possible to actually consolidate anything, it was possible to differentiate and fix some things:

- Split `getFullDimensionId` into `joinDimensionIdParts` (pure mechanical inverse of `getDimensionIdParts`) and `composeAnalyticsRequestHeader` (the analytics-request-specific rules for `created` and standalone enrollment-scoped tokens). The old name implied a canonical-ID builder but mixed structural joining with request-side rules.
- Added `composeLineListColumnId` to replace the response-header → app-local-id logic inlined three times across `use-transformed-line-list-data.ts` and `use-line-list-analytics-data.ts`.
- Renamed `getCompoundDimensionId` → `getCanonicalDimensionId` since the result is sometimes plain, not compound (program indicators, TEAs).
- Line-list call sites no longer roundtrip through `headersMap` to recover the canonical key just to compare against it — they match the lowercase header token directly.
- Updated CLAUDE.md to reference the new names and the correct loader entry point (`getVisualizationUiConfig`).

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

[DHIS2-21468]: https://dhis2.atlassian.net/browse/DHIS2-21468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/plugin-wrapper/pivot-table-plugin.tsx
```diff
@@ -33,7 +33,7 @@ const formatVisualizationForPivotTableEngine = (
 
             return {
                 ...dimensionObj,
-                dimension: getFullDimensionId({
+                dimension: composeAnalyticsRequestHeader({
```
edoardo, line 36: The name here is a bit confusing since PT does not use headers.

src/modules/visualization.ts
```diff
@@ -83,6 +84,106 @@ export const getHeadersMap = (
...
+ * Composes the analytics-request column header for an app-side dimension.
+ *
+ * Pass the *lowercase header token* (the value from `headersMap`) as
+ * `dimensionId` — UIDs and tokens not in `headersMap` are passed through
+ * unchanged. The helper joins program/stage prefixes via `joinDimensionIdParts`
+ * except for the standalone tokens above (and `created`, which is always
+ * bare since it lives at TEI-registration scope and the analytics API never
+ * accepts it prefixed).
+ */
+export const composeAnalyticsRequestHeader = ({
```
edoardo, line 111: I think I would prefer renaming this to something that does not mention header, since this is also used for PT where the concept of header is foreign.

edoardo (commented):
This PR touches some of the same files as #209 so it's a bit difficult to follow since both are merging into master.

PR #224, opened 2026-05-12
chore: remove unused initial metadata

HendrikThePendric:
Implements N/A

### Description

While working on some other tasks it became apparent to me that some of the initial metadata has keys that do not match the app-local compound IDs. This metadata with the correct keys is produced while extracting the metadata from the visualization. So this initial metadata is now useless and can be removed.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### DIFF
--- a/src/components/app-wrapper/metadata-provider/__tests__/metadata-store.spec.ts
+++ b/src/components/app-wrapper/metadata-provider/__tests__/metadata-store.spec.ts
@@ -433,18 +433,6 @@ describe('MetadataStore', () => {
               "id": "enrollmentDate",
               "name": "Enrollment date",
             },
-            "eventDate": {
-              "dimensionId": "eventDate",
-              "dimensionType": "PERIOD",
-              "id": "eventDate",
-              "name": "Event date",
-            },
-            "eventStatus": {
-              "dimensionId": "eventStatus",
-              "dimensionType": "STATUS",
-              "id": "eventStatus",
-              "name": "Event status",
-            },
             "gWxh7DiRmG7": {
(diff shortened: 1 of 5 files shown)


src/components/app-wrapper/metadata-provider/__tests__/metadata-store.spec.ts
```diff
@@ -602,6 +578,12 @@ describe('MetadataStore', () => {
             )
         ).toMatchInlineSnapshot(`
           [
+            "completedDate",
+            "createdDate",
```
edoardo, line 582:
```suggestion
            "completed",
            "created",
```

I'm confused about these 2.
Are they ever passed/saved as `createdDate` and `completedDate`?

edoardo (changes requested):
This looks fine, except for the commented code.
It's only in a test, so not that important, but there are references to those dimensions in the code.

src/components/app-wrapper/metadata-provider/__tests__/metadata-store.spec.ts
edoardo: This will be addressed in a separate PR as it needs clarification and related changes in different files.

edoardo (approved):

src/components/app-wrapper/metadata-provider/__tests__/metadata-store.spec.ts
HendrikThePendric: See [DHIS2-21504](https://dhis2.atlassian.net/browse/DHIS2-21504)

PR #229, opened 2026-05-18
feat: sidebar disabling [DHIS2-21238]

HendrikThePendric:
Implements [DHIS2-21238](https://dhis2.atlassian.net/browse/DHIS2-21238)

### Description

Disables sidebar cards and dimension items that aren't valid for the current visualization config — dims them, surfaces a NoticeBox above the first disabled card (or a tooltip on disabled dim items) explaining why.

Scope is the sidebar only. Conversion, drag-and-drop and the conditions modal are handled separately.

Spec: [Invalid Dimension Strategy](https://docs.google.com/document/d/1TQ6hEjXRoSS-v4_OHnEgZA6MuVgUO8D0EjcV1wvrnFM/edit?usp=sharing)

#### Changes to application logic

Two pre-existing assumptions surfaced while implementing the disabling rules and needed fixing:

1. **\`resolveTetId\` (\`src/modules/layout.ts\`) now derives the layout's TET from any TET-bound dim, not just the OU.** A valid layout can only reference a single TET, so any TET-bound dimension is a sufficient anchor — there's no reason to require the OU specifically. Side benefit: the TEI action button now correctly detects the TET when the layout contains only TEAs and no OU.

2. **TEAs now use compound \`tetId.attrId\` IDs in app-local state.** The backend allows an attribute UID to be referenced by multiple TETs, but in the frontend the same UID reached via two TETs is functionally two separate sidebar items. Prefixing with \`tetId\` disambiguates them so per-card \`selectedCount\` and \`useIsDimensionInLayout\` can't collide. Applied at the load boundary (\`getCompoundDimensionId\`) and stripped at the save boundary (\`buildAxis\`), so the wire format is unchanged.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A
- [x] d2-ci dependency replaced

---

### Screenshots

_supporting images_

[DHIS2-21238]: https://dhis2.atlassian.net/browse/DHIS2-21238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edoardo (dismissed):
One general comment is about the use of `vizType`.
I don't think we used that before.
I would change that to `visType` which we already use.

edoardo (approved):

PR #235, opened 2026-05-19
fix: strip program/programStage from all contextless dimensions

edoardo:
### Description

Contextless dimensions are the one that should be passed to the analytics request without a prefix (program/programStage).
The list was already present in the code, it has been moved to `modules/dimension` and used in `toAppLocalDimensions` to strip the program/programStage information from the dimension object.
Now the analytics request is computed correctly.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

Before:
Screenshot 2026-05-19 at 16 18 32
Screenshot 2026-05-19 at 16 18 40

After:
Screenshot 2026-05-19 at 16 19 04
Screenshot 2026-05-19 at 16 18 53

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2026-05-18T13:11:36.297Z\n"
-"PO-Revision-Date: 2026-05-18T13:11:36.297Z\n"
+"POT-Creation-Date: 2026-05-19T14:03:47.193Z\n"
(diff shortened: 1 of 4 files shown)


HendrikThePendric (changes requested):
Maybe the naming you applied here makes sense for reasons I am failing to see, but I was expecting something like this:
- When you moved `getFixedDimensions` from the metadata card, it should have been renamed to `getFixedMetaDimensions` 
- For consistency `CONTEXTLESS_DIMENSION_IDS` should be `META_DIMENSION_IDS`

Note I used `metaDimensions` instead of `metadataDimensions` to differentiate between:
- Metadata from the metadata store
- Dimensions in the the metadata card

EDIT: My main objection to the current state of the codebase is the overly generic name `getFixedDimensions`. Each card uses a mix of fixed and/or fetched dimensions, so what a fixed dimension is depends per card.

HendrikThePendric (approved):

PR #240, opened 2026-05-21
feat: custom value compound id [DHIS2-21538]

HendrikThePendric:
Implements [DHIS2-21538](https://dhis2.atlassian.net/browse/DHIS2-21538) and some scenarios we uncovered along the way

#### Compound dimension IDs

   We're now using a different endpoint to populate the list of dimensions in the custom value modal. This endpoint returns compound IDs in `stageId.dimensionId` form, which means we send the correct ID format when saving a visualisation or doing analytics requests, and — more visibly to the user — the sidebar can disable the exact dimension that is currently in use as the custom value. Disabling matches by the compound ID, so a different stage's instance of the same data element stays enabled, with a tooltip explaining why a dimension is disabled. As part of the same disabling pass, the TET registration OU is also disabled when the visualization type is `PIVOT_TABLE` (the backend doesn't support that combination).

#### Metadata store auto-derives stages and TET

   While working on this I realised we had quite a few occurrences where we add a program to the metadata store. When doing so we then would always need to manually extract the stages and the TET and add these to the store as well. I encountered a place where we were not doing this and that was causing the code in the custom-value-modal to become quite complex. So what I did instead is add a feature to the metadata store: when you pass a program as input the metadata store will create entries for the program, the stages and the TET. This makes things a lot simpler in the rest of the app.

#### Custom value button refactor

   The split button is now rendered directly inside `CustomValueButton` — the `BaseCustomValueButton` indirection is gone. Added a small `WithTooltip` helper that conditionally wraps its children so the two halves of the split button can carry different tooltips. The update half is now disabled both when the action itself is disabled and when there is a stage mismatch, and the relevant tooltip is shown for each case. The "using …" tooltip on the configure half is unchanged.

#### Stage-aware custom value modal

   Moved the modal out of `bottom-bar/` into its own `custom-value-modal/` folder and split it up:
    - A new `useCustomValueDataElements` hook fetches the dimensions, scopes them to the layout's program stage if there is one, and reports whether the currently saved custom value is from a different stage.
    - A new `StageNotice` shows either an info banner ("Showing data elements from stage …") or a warning when the saved custom value belongs to a different stage and prompts the user to pick a new data element.
    - A new `CustomValueOption` renders each option with an optional stage chip — used when the layout has no stage filter and the list spans multiple stages.
    - Unit tests added for the modal, the option component, and the data-elements hook.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

Screenshot 2026-05-26 at 15 58 26

_Note how "E2E TE P1 - DE 1 (cross stage)" is disabled for stage 1 but not disabled under stage 2_

https://github.com/user-attachments/assets/47f53e1e-c231-4ae9-9d9e-2ec358e3bd90

[DHIS2-21538]: https://dhis2.atlassian.net/browse/DHIS2-21538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/layout-panel/custom-value-modal/stage-notice.tsx
```diff
@@ -0,0 +1,43 @@
...
+                        {
+                            dataElementName: customValueDataElementName ?? '',
+                            stageName: filteredByStageName,
+                        }
+                    )}
+                </NoticeBox>
+            ) : (
+                <NoticeBox
+                    title={i18n.t(
+                        'Showing data elements from stage "{{- stageName}}" which is used in the layout',
```
cooper-joe:
```suggestion
                        'Showing data elements from "{{- stageName}}", the stage used in the layout',
```

src/components/layout-panel/bottom-bar/action-buttons/custom-value-button.tsx
```diff
@@ -1,93 +1,93 @@
...
+                  dataElementName: customValueMetadata?.name,
+                  aggregationType:
+                      aggregationTypeDisplayNames[customValue.aggregationType],
+                  nsSeparator: '^^',
+              })
+            : undefined
+    const stageMismatchTooltipContent =
+        !isFullyDisabled && hasStageMismatch
+            ? i18n.t(
+                  'Currently selected custom value data element is from a different stage than the dimensions in the layout'
```
cooper-joe:
```suggestion
                  'Custom value is from a different stage than dimensions in the layout'
```

src/components/layout-panel/bottom-bar/action-buttons/custom-value-button.tsx
```diff
@@ -1,93 +1,93 @@
...
+    const isUpdateDisabled = isFullyDisabled || hasStageMismatch
+    const configurationTooltipContent =
+        !isFullyDisabled && customValue && !hasStageMismatch
+            ? i18n.t('Using: {{- dataElementName}} ({{- aggregationType}})', {
+                  dataElementName: customValueMetadata?.name,
+                  aggregationType:
+                      aggregationTypeDisplayNames[customValue.aggregationType],
+                  nsSeparator: '^^',
+              })
+            : undefined
```
cooper-joe: In this state, tooltip content of `Configure custom value` is probably helpful and worth having.

src/components/layout-panel/custom-value-modal/stage-notice.tsx
```diff
@@ -0,0 +1,43 @@
...
+    if (!filteredByStageName) {
+        return null
+    }
+
+    return (
+        <div className={classes.stageNotice}>
+            {customValueStageMismatch ? (
+                <NoticeBox warning>
+                    {i18n.t(
+                        'The currently selected data element "{{- dataElementName}}" is from a different stage than the dimensions in the layout. Pick a new data element from stage "{{- stageName}}".',
```
cooper-joe:
```suggestion
                        '"{{- dataElementName}}" is from a different stage than the dimensions in the layout. Choose a data element from stage "{{- stageName}}".',
```

cooper-joe (changes requested):
Image

The custom value list shows the `stageName` chip even if there is only one stage in the program. We can omit the chip in this case.

---

Image

"No numeric data items in this program" is not true, it is the filtered stage in this case. We need a stage-aware variant of this `NoticeBox`, with content "No numeric data items in stage $stageName".

---

Let's use the `NoticeBox dense` variant for all notice boxes in the Custom Value modal.

src/components/layout-panel/custom-value-modal/use-custom-value-data-elements.ts
```diff
@@ -0,0 +1,123 @@
...
+    const { data, ...queryResult } = useRtkQuery<{
+        dimensions: DataElementDimension[]
+    }>({
+        resource: 'analytics/enrollments/aggregate/dimensions',
+        params: {
+            programId,
+            fields: `id,${displayNameProperty}~rename(name),aggregationType,dimensionType`,
+            filter: [
+                'dimensionType:eq:DATA_ELEMENT',
+                // TODO: add BOOLEAN and TRUE_ONLY if/when backend supports it
```
edoardo: Did you try enabling these in this endpoint?

edoardo (dismissed):

src/components/layout-panel/custom-value-modal/use-custom-value-data-elements.ts
HendrikThePendric: As discussed, I have added these two, see https://github.com/dhis2/event-visualizer-app/pull/240/commits/3ae6eb0e0b99019285d3f099005e908c8df02145

src/components/layout-panel/bottom-bar/action-buttons/custom-value-button.tsx
HendrikThePendric: I actually went for "Update custom value" here to signal the _need_ to change it.

cooper-joe (approved):

PR #241, opened 2026-05-22
fix: fix download links for both LL and PT

edoardo:
### Description

Download links didn't work properly after the various changes in the analytics requests.
This PR adds fixes to the LL downloads and implements the PT download which was not in place.

For PT the "Advanced" sub-menu is also available, similar to ER/DV.

We need to verify the various downloads, especially the ones for PT, since there are differences in the way we build the analytics request with the Event Reports app.

There is also a fix for `programStatus` which was ignored when present in the vis obj, but needs to be taken into account and converted into a dimension in filters.

The PR also has some fixes to avoid a situation where the `currentVis` is updated but the previous analytics data is used when changing layout/outputType and clicking one of the action buttons.
This was especially problematic in PT where the PT engine would throw an error due to missing metadata.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### ToDos

- [x] verify the various downloads (this will have to be done later)
- [x] check with Jan if XLS and CSV are needed in the top section (table download type)

### Screenshots

Download menu for LL:
Screenshot 2026-05-29 at 16 32 04

Download menu for PT:
Screenshot 2026-05-29 at 16 31 53


src/components/download-menu/advanced-sub-menu.tsx
HendrikThePendric, line 1: Just wondering about the name here. Will this remain specific for the Pivot Table, or will charts use this too? 

I am guessing that charts need a different menu, and if so, then I would prefer making this name less generic, and perhaps call it sth like `PivotTableSubMenu`

src/components/download-menu/use-download.ts
```diff
@@ -166,11 +263,14 @@ const useDownload: (relativePeriodDate?: string) => UseDownloadResult = (
     )
 
     // XXX: should the transformed visualization be passed here?
```
HendrikThePendric: Comment can be removed. We use CurrentVis and that's good

src/components/plugin-wrapper/hooks/use-line-list-analytics-data.ts
```diff
@@ -327,6 +329,12 @@ const useLineListAnalyticsData = (): UseAnalyticsDataResult => {
         data: null,
     })
 
+    const [prevLayoutKey, setPrevLayoutKey] = useState(layoutKey)
+    if (prevLayoutKey !== layoutKey) {
+        setPrevLayoutKey(layoutKey)
+        setState({ isFetching: false, error: undefined, data: null })
+    }
```
HendrikThePendric: I'm thinking about this part of code and wondering if this is the most idiomatic way to do this... You're simply resetting the state when the `layoutKey` changes right?

To solve this we have some other solutions available too:
1. Add `key={layoutKey}` to the component that consumes the hook, this will cause a remount on change. Might be too aggressive
2. Use `useEffect` with `layoutKey` as dependency and reset there

src/components/plugin-wrapper/line-list-plugin.tsx
```diff
@@ -31,8 +31,20 @@ export const LineListPlugin: FC<LineListPluginProps> = ({
...
+        () =>
+            `${visualization.outputType}|${[
+                ...visualization.columns,
+                ...visualization.rows,
+                ...visualization.filters,
+            ]
+                .map((d) => d.dimension)
+                .join('|')}`,
+        [visualization]
+    )
```
HendrikThePendric: Aha. The `layoutKey` is computed in the consuming component, which makes solution 1 in my previous suggestion invalid. Now that I see the full picture I'd recommend the following:
1. Pass `visualization` to the hook
2. Extract a `computeLayoutKey` helper (so this can be reused)
3. Have a memoized `layoutKey` in the hook(s)
4. Use useEffect with dependency on `layoutKey` to reset

HendrikThePendric (commented):
I've left a few suggestions. Nothing critical though.

src/components/download-menu/advanced-sub-menu.tsx
edoardo: If we compare the download menu with DV and EV, charts also have this same Advanced submenu.
What's different is the top section where you will have image/pdf downloads.
I think I'll leave as is for now.

src/components/plugin-wrapper/line-list-plugin.tsx
edoardo: I've refactored as you suggested except for point 4.
Here's Claude's reasoning:
"The derived state reset is kept as-is — the useEffect approach would reintroduce the crash since effects fire after render, meaning the PT engine would get one frame of new visualization + stale data before
  the reset could run."

edoardo: We agreed on refactor and simplify this.
Now a memoized selector in the `currentVis` slice is used to computed the layout key based on layout + outputType.
This key is then passed as props to the plugins.

src/store/current-vis-slice.ts
```diff
@@ -54,3 +54,14 @@ export const currentVisSlice = createSlice({
...
+export const getCurrentVisLayoutKey = createSelector(getCurrentVis, (vis) =>
+    [
+        vis.outputType ?? '',
+        ...[
+            ...(vis.columns ?? []),
+            ...(vis.rows ?? []),
+            ...(vis.filters ?? []),
+        ].map((d) => d.dimension),
+    ].join('|')
+)
```
HendrikThePendric: I think this can memoized a bit smarter. The inputs should be all the parts of the visualization we are interested in, not the full vis object.

Also, since this is a "pure" memoized selector (it's all based on redux state, no external arguments are provided), this selector would also be part of the selectors in the slice. But TBH I am still not completely sure what the proper place for these selectors is.

🤔 Actually, now that I am thinking about it, we possibly do have an established pattern for this, provided that this selector is only used by the plugin-wrapper component. If so, just colocate it with that component.

HendrikThePendric (dismissed):
Small, suggestion but nothing critical

HendrikThePendric (approved):

PR #252, opened 2026-06-01
chore: clean up old dimension names and normalize [DHIS2-21504]

HendrikThePendric:
Implements [DHIS2-21504](https://dhis2.atlassian.net/browse/DHIS2-21504)

### Description

The app used three old dimension IDs — `createdDate`, `completedDate` and `lastUpdatedOn` — as time dimensions. These were redundant duplicates of the real metadata dimensions `created`, `completed` and `lastUpdated`. The analytics API and the line-listing-app both use the canonical names, so the old ones were removed and we now only generate `created`/`completed`/`lastUpdated`.

We still normalize on load because saved visualizations from the old event-visualizer (Event Reports) app can contain the old IDs (`createdDate` is a real alias). When one loads, `normalizeApiSavedVisualization` rewrites the old IDs to the new ones so the app can render them.

The API accepts some old names as aliases, but we still mark any rewritten visualization as **legacy** (only "Save as" enabled): the old app may rely on the old names (e.g. for labels), so we never overwrite it in place. The rule is now simple — if loading changed anything, the visualization is legacy. 

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

[DHIS2-21504]: https://dhis2.atlassian.net/browse/DHIS2-21504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/dimension.ts
+++ b/src/modules/dimension.ts
@@ -30,6 +30,7 @@ export const outputTypeTimeDimensionMap: Record<OutputType, DimensionId> = {
 export const timeFieldTimeDimensionMap: Record<string, DimensionId> = {
     COMPLETED_DATE: 'completed',
     CREATED: 'created',
+    CREATED_DATE: 'created',
     ENROLLMENT_DATE: 'enrollmentDate',
     EVENT_DATE: 'eventDate',
     INCIDENT_DATE: 'incidentDate',
--- a/src/modules/dimension.ts
+++ b/src/modules/dimension.ts
@@ -28,7 +28,7 @@ export const outputTypeTimeDimensionMap: Record<OutputType, DimensionId> = {
  * concrete time-dimension id the app uses internally. Used when
  * materialising a legacy `pe` dimension into a proper time dimension. */
 export const timeFieldTimeDimensionMap: Record<string, DimensionId> = {
-    COMPLETED_DATE: 'completedDate',
+    COMPLETED_DATE: 'completed',
     CREATED: 'created',
     ENROLLMENT_DATE: 'enrollmentDate',
     EVENT_DATE: 'eventDate',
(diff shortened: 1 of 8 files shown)


src/modules/dimension.ts
```diff
@@ -28,7 +28,7 @@ export const outputTypeTimeDimensionMap: Record<OutputType, DimensionId> = {
  * concrete time-dimension id the app uses internally. Used when
  * materialising a legacy `pe` dimension into a proper time dimension. */
 export const timeFieldTimeDimensionMap: Record<string, DimensionId> = {
-    COMPLETED_DATE: 'completedDate',
+    COMPLETED_DATE: 'completed',
     CREATED: 'created',
```
edoardo, line 32: Should we use the same format for these keys? ie. `CREATED_DATE` for `created`?

edoardo (dismissed):

src/modules/dimension.ts
HendrikThePendric: Quick background on what this map is for:  It only matters for legacy visualizations. Older line-listing visualizations didn't save a proper date dimension — they saved a  generic period dimension (`pe`) plus a separate `timeField` value saying which date that period meant. `timeField` is the same name the analytics API uses as a query parameter for picking which date to filter on; here we read it off the saved visualization when loading it. This map is how we turn that old `pe` + `timeField` combo into the real date dimension the app uses today. Modern visualizations save the date dimension directly and never touch this.

On the format: the keys aren't ours to pick — they're the exact values the backend's `TimeField` enum produces, so they have to match its spelling (which is why `LAST_UPDATED` has no `_DATE` either).

On `CREATED` vs `CREATED_DATE`: these are two different values in `TimeField`, pointing at different database columns:
- `CREATED(CREATED_COLUMN_NAME)`
- `CREATED_DATE(CREATED_DATE_COLUMN_NAME)`

A legacy visualization could have saved either string in its `timeField`. Both mean a "created" timestamp, so both should resolve to the app's created dimension. We were only handling `CREATED`, so a vis that happened to save `CREATED_DATE` would silently fail to convert. I've added `CREATED_DATE: 'created'` so both cases are covered.

edoardo (approved):

PR #257, opened 2026-06-03
fix: ui fixes and styling

cooper-joe:
### Description

This PR is a batch of UX/UI tweaks and fixes. Aside from the mostly css changes, there are a few functionality changes:
- **Custom value list sorting**: data element list is sorted A-Z by name (previously was ordered arbitrarily from API)
- **Prevent sidebar auto-scroll while dragging**: configured the dnd-kit `autoScrollOptions` to stop the sidebar scrolling while dragging a dimension.
- **Empty chip styling**: applied the empty styles directly to `ChipBase` and cleaned up some dead styles not being applied.

---

### Quality checklist

- [x] Dashboard tested (N/A)
- [x] Docs added (N/A)
- [x] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/XXX>) (N/A)


src/components/layout-panel/custom-value-modal/use-custom-value-data-elements.ts
```diff
@@ -100,25 +100,32 @@ export const useCustomValueDataElements = () => {
             return undefined
         }
 
+        const byName = (a: { name: string }, b: { name: string }) =>
+            a.name.localeCompare(b.name)
```
HendrikThePendric: This comparator can be declared outside of the function scope. And I don't think we need a new type for this? These are `DataElementDimension`s right?

src/components/app-wrapper/drag-and-drop-provider/dnd-context-provider.tsx
```diff
@@ -10,6 +16,12 @@ const activateAt15pixels = {
     },
 }
 
+const DISABLE_AUTO_SCROLL_SELECTOR = '[data-dnd-auto-scroll="disabled"]'
+
+const autoScrollOptions: AutoScrollOptions = {
+    canScroll: (element) => !element.matches(DISABLE_AUTO_SCROLL_SELECTOR),
+}
```
HendrikThePendric, line 23: I think/hope we would be able to resolve this in a different way.

Options I am considering:
- Can we not just set a prop on the sidebar SortableProvider?
- A fix similar to this but not using a data-attribute, but using the drag data instead.

src/components/sidebar/sidebar.tsx
```diff
@@ -58,6 +58,7 @@ export const Sidebar: FC = () => {
                     <div
                         onScroll={handleScroll}
                         className={classes.dimensionCardsContainer}
+                        data-dnd-auto-scroll="disabled"
```
HendrikThePendric, line 61: See my previous suggestion

HendrikThePendric (changes requested):
Looking good, but would like to try some minor changes, see comments.

src/components/app-wrapper/drag-and-drop-provider/dnd-context-provider.tsx
HendrikThePendric: OK, apparently this is _the DnD Kit way_ to resolve these things. Slightly ugly but it works.

src/components/sidebar/sidebar.tsx
HendrikThePendric: See https://github.com/dhis2/event-visualizer-app/pull/257#discussion_r3373399049

HendrikThePendric (approved):

PR #264, opened 2026-06-08
feat: update button icon animation

cooper-joe:
### Description

This PR animates the _Update_ button icon on click. The animation is also triggered when updating via a modal dialog or when switching to another output type. The animation does not play when creating a visualisation for the first time as this felt wrong.
(This feature is in response to user feedback where there was an underlying feeling of "did anything happen"?)

Note: The animation does not play when updating from a custom value modal dialog, but that is a bug with that dialog, not in the animation. Fix for that should outside the scope of this PR.

#### How it works
`updateAnimationShowingFor: OutputType | null` records which button should be animating. The update thunk sets it to the current output type on every update (except on first create, when the visualization is empty). Each update button reads the value and passes `isAnimating` as a boolean prop to its icon component.

The icon component is a plain SVG that applies a `.run` CSS class when `isAnimating` is true, which triggers a `@keyframes` animation defined in its module CSS. A `useUpdateAnimation` hook runs 500ms timer that clears the store value once the animation completes, which drops the class and leaves the icon idle.

All update paths (layout buttons and dimension/option modals) go through the same thunk, so the bottom bar icon responds regardless of where the update originates.

---

### Quality checklist

- [x] Dashboard tested _N/A_
- [x] Docs added _N/A_
- [x] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/XXX>)

### Screenshots

https://github.com/user-attachments/assets/6d4e786f-f8c1-4a31-ae08-e1dddead71d9


HendrikThePendric (changes requested):
The diff is looking great, but I think there is a small but fundamental problem here. The new field in the state-slice looks like this:
```ts
interface PanelVisibility {
    updateAnimationShowingFor: OutputType | null
}
```

But in some cases `OutputType` is not enough to identify the button, because when the `visType` is `PIVOT_TABLE` the `outputType` for both the event-list as well as the custom-value-list button is `EVENT`, this would mean both buttons would show the animation.

Currently there is no elegant way to differentiate between these two, but in #269 I have made some changes that will actually make this possible. So I suggest we consider this PR blocked by that one and once that's merged, we can add some code to `useUpdateAnimation` that infers the active button type from the current vis:
- If the `currentVis` has a `value` and `aggregationType` field then the active button is the custom-value-list button
- If not it is the event-list button
- One thing to keep in mind is timing though. We need to ensure we first update the `currentVis` and then we update the `updateAnimationShowingFor`, because if we mess up the order we could be reading a bad state snapshot

HendrikThePendric (dismissed):
I pushed a commit to address the concern I raised in my previous review. It adds some code to differentiate between the even VS custom value update buttons. There is nothing wrong with it, the logic is sound.

But after I produced the fix I also realised that at present there technically is no need to differentiate because we conditionally render the icon and it's only visible in one of the buttons at a time.

So we can keep my commit or drop it. Either way, this PR is approved.

HendrikThePendric (approved):

PR #269, opened 2026-06-10
chore: switching between custom value and event tables

HendrikThePendric:
Implements various things:
- demo blocking issue "Switching from custom value table to event table"
- [DHIS2-21279](https://dhis2.atlassian.net/browse/DHIS2-21279)
- Some addition improvements

### Description

- Switching between custom value and event list now works:
    - The custom value `value` field is now cleared from the current vis when switching to a different `outputType`, or from custom-value-list to event-list (technically the same `outputType`). The remembered custom value selection is kept in `visUiConfig`, so switching back reuses it without reopening the modal.
    - The custom values fields are now taken into account when computing the `currentVisLayoutKey`. This causes the event-list and custom-value-list to have a different key
- The above gave us a way to determine the last active button from the `currentVis` so we were able to remove the `lastActiveButton` state completely ([DHIS2-21279](https://dhis2.atlassian.net/browse/DHIS2-21279))
- [BONUS] Added a search field above the data-elements-list in the custom-value-modal
- [BONUS] Fixed focus styles on the the data-elements-list-items in the custom-value-modal
- [BONUS] Clicking Update" in the custom value modal now both updates the custom value config AND it updates the pivot-table

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### ToDos

- [x] Get approval on bonus-features (approved by @cooper-joe on Slack)

---

### Screenshots

Screenshot 2026-06-10 at 10 54 42

[DHIS2-21279]: https://dhis2.atlassian.net/browse/DHIS2-21279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-21279]: https://dhis2.atlassian.net/browse/DHIS2-21279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout-panel/bottom-bar/action-buttons/base-button.tsx
+++ b/src/components/layout-panel/bottom-bar/action-buttons/base-button.tsx
@@ -30,7 +30,7 @@ const BaseButton: FC<BaseButtonProps> = ({
 
     const onClick = () => {
         dispatch(setVisUiConfigOutputType(type))
-        dispatch(tUpdateCurrentVisFromVisUiConfig('EVENT'))
+        dispatch(tUpdateCurrentVisFromVisUiConfig(false))
     }
 
     return (
--- a/src/components/layout-panel/bottom-bar/action-buttons/base-button.tsx
+++ b/src/components/layout-panel/bottom-bar/action-buttons/base-button.tsx
@@ -1,11 +1,7 @@
 import { IconSync16, Tooltip } from '@dhis2/ui'
 import { useAppDispatch } from '@hooks'
 import { tUpdateCurrentVisFromVisUiConfig } from '@store/thunks'
-import {
-    setVisUiConfigLastActiveButton,
-    setVisUiConfigOutputType,
-} from '@store/vis-ui-config-slice'
-import type { LastActiveButton } from '@store/vis-ui-config-slice'
+import { setVisUiConfigOutputType } from '@store/vis-ui-config-slice'
 import type { OutputType } from '@types'
 import cx from 'classnames'
 import { type FC } from 'react'
@@ -18,7 +14,6 @@ export type BaseButtonProps = {
     dataTest?: string
     disabled?: boolean
     label: string
-    lastActiveButton?: LastActiveButton
     tooltipProps?: object
     type: OutputType
 }
@@ -28,18 +23,14 @@ const BaseButton: FC<BaseButtonProps> = ({
     dataTest,
     disabled = false,
     label,
-    lastActiveButton,
     tooltipProps,
     type,
 }) => {
     const dispatch = useAppDispatch()
 
     const onClick = () => {
-        if (lastActiveButton !== undefined) {
-            dispatch(setVisUiConfigLastActiveButton(lastActiveButton))
-        }
         dispatch(setVisUiConfigOutputType(type))
-        dispatch(tUpdateCurrentVisFromVisUiConfig())
+        dispatch(tUpdateCurrentVisFromVisUiConfig('EVENT'))
     }
 
     return (
--- a/src/store/thunks.ts
+++ b/src/store/thunks.ts
@@ -40,6 +40,8 @@ type AppThunk = () => (
     extra: ThunkExtraArg
 ) => void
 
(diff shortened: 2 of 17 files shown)


src/components/layout-panel/bottom-bar/action-buttons/base-button.tsx
```diff
@@ -28,18 +23,14 @@ const BaseButton: FC<BaseButtonProps> = ({
...
 }) => {
     const dispatch = useAppDispatch()
 
     const onClick = () => {
-        if (lastActiveButton !== undefined) {
-            dispatch(setVisUiConfigLastActiveButton(lastActiveButton))
-        }
         dispatch(setVisUiConfigOutputType(type))
-        dispatch(tUpdateCurrentVisFromVisUiConfig())
+        dispatch(tUpdateCurrentVisFromVisUiConfig('EVENT'))
```
edoardo: I'm confused about this one. What about the ENROLLMENT button?
I understand in that case EVENT prevents the customValue stuff from being stored in `currentVis` and `outputType` is not read from this parameter anyway, but it's a bit confusing seeing the `EVENT` variant passed for the enrollment button.

src/store/thunks.ts
```diff
@@ -104,15 +106,32 @@ export const tLoadSavedVisualization = createAsyncThunk<
...
+    ) => {
         const state = getState()
         const { currentVis, visUiConfig } = state
         const { metadataStore } = extra
         const { customValue } = visUiConfig
 
+        const resolvedVariant: EventOutputTypeVariant =
+            variant ?? (currentVis.value?.id ? 'CUSTOM_VALUE' : 'EVENT')
+        const customValueFields =
+            resolvedVariant === 'CUSTOM_VALUE' && customValue
```
edoardo: Should we add a check on `outputType === 'EVENT'` here so we can avoid passing `EVENT` to the thunk by default?

edoardo (commented):
Added comments about the variant usage in the thunk.

src/components/layout-panel/bottom-bar/action-buttons/base-button.tsx
HendrikThePendric: Thanks for flagging this up. The confusion is completely justified. I've refactored the implementation to hopefully make things a bit clearer, but I'll also explain the broader context here. The challenge is that we're effectively dealing with three cases:
1. The custom value button in the layout and custom value button in the custom value modal. These need to explicitly signal: add the custom value fields to the `currentVis`
2. The other update buttons in the layout. The need to explicitly signal: remove the custom value fields from the `currentVis`
3. The update buttons in all dimension modals, which need to signal: leave the custom value fields in whatever state they were before

To reduce the confusion we now pass a boolean to the thunk where `true` means add, `false` means remove, and `undefined` means keep. In the thunk this boolean is called `withCustomValue`.

src/store/thunks.ts
HendrikThePendric: This comment should no longer be relevant after the refactor I did.

src/store/thunks.ts
```diff
@@ -106,8 +104,48 @@ export const tLoadSavedVisualization = createAsyncThunk<
...
+}
+
+const resolveCustomValueFields = (
+    state: RootState,
+    withCustomValue?: boolean
+) => {
+    // Always include the `value` key: setCurrentVis merges into the previous
+    // currentVis, so omitting it would leave a stale value behind.
+    if (!shouldShowCustomValue(state, withCustomValue)) {
+        return { value: undefined }
```
edoardo: You asked me about this, and I think `aggregationType` should also be cleared.

edoardo (dismissed):

src/store/thunks.ts
HendrikThePendric: Done

edoardo (approved):

PR #275, opened 2026-06-11
fix: various ui fixes

cooper-joe:
### Description

This PR implements various UI fixes:
- prevents a 1px jump in the toolbar when empty vs. populated
- prevents an _Update_ action from the Options dialog when the visualisation is empty
- uses a stable scrollbar in the layout so that dimensions don't shift when resizing vertically
- adjust empty state label

---

### Quality checklist

- [x] Dashboard tested
- [x] Docs added
- [x] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/XXX>)


HendrikThePendric (approved):
LGTM

PR #277, opened 2026-06-11
chore: enable TypeScript full strict mode

HendrikThePendric:
Implements [DHIS2-21614](https://dhis2.atlassian.net/browse/DHIS2-21614)

### Description

Stage 2 of the TS 6.0 upgrade: enable `strict: true` in both `tsconfig.json` and `cypress/tsconfig.json` and fix everything it surfaces.

**Type-only fixes**

- The bulk: `noImplicitAny` params/indexing, `strictFunctionTypes` reducer-variance in test-store factories (pin `ReducersMapObject`'s PreloadedState generic to `Partial<RootState>`), a `strictNullChecks` guard in `save.cy.ts`.
- Correct two inaccurate `@dhis2/analytics` shims (`visTypeDisplayNames` is a `Record` not an `Array`; `PeriodDimension.onSelect` items are objects, not strings).

**Runtime bugs surfaced during review**

- Pivot-table `sortOrder` always picked `DESC` because of a `number === '1'` comparison that was always false.
- `getHeadersMap` was being called with `getRequestOptions(visualization)` — a stripped options object missing `outputType` / `type` / `showHierarchy`, silently breaking the `PIVOT_TABLE` / `ENROLLMENT` / hierarchy branches. (Pre-existing bug, opportunistically fixed.)
- Analytics boolean and status cells fell silently to `undefined` for unknown enum values; now fall back to the raw value.

**Type model improvements**

- `OrgUnit.id` tightened to required; `TimeDimension.nameProperty` constrained to a discriminated union of actual label keys; `CachedDataQueryProvider.dataTransformation` typed generically.
- New `useDimensionMetadataItems` hook for typed plural lookup. Several `as unknown as X` casts replaced with type predicates (`isStatus`, `isBooleanValue`) and honest type signatures.

---

### Quality checklist

- [x] Dashboard tested — N/A
- [x] Docs added N/A

[DHIS2-21614]: https://dhis2.atlassian.net/browse/DHIS2-21614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/plugin-wrapper/hooks/use-pivot-table-analytics-data.ts
edoardo, line 42: Shouldn't this be the following then?
```suggestion
            parameters: mutableParameters,
```

edoardo (changes requested):

edoardo (approved):

PR #287, opened 2026-06-16
chore: put Claude in a sandbox

HendrikThePendric:
Implements: No JIRA ticket available

### Description

An **opt-in** pair of AI dev sandboxes built on [Docker Sandboxes](https://docs.docker.com/ai/sandboxes/) (`sbx`): a hands-on **mount** (works on your live files) and an autonomous **clone**. If you don't install `sbx`, none of it runs. Full guide: [`docs/claude-sandboxes.md`](docs/claude-sandboxes.md). Highly experimental — tested only on macOS/arm64 + Neovim.

What it provides: isolated workspaces with no permission prompts, in-sandbox tooling (pnpm, plugins, grep MCP, format hook), IDE integration, headless-browser automation, a native-filesystem `node_modules` overlay for fast tests, and the `sonarqube-fix` skill in the mount (if you have a SonarCloud token).

**Impact on the regular (non-sandbox) host workflow** — it's opt-in, but a few committed files affect everyone using Claude Code on this repo:

| Change | Effect on host devs |
| --- | --- |
| Format hook → `scripts/claude-format-hook.sh` | ✅ Was a silent no-op (read a non-existent env var); now actually auto-formats + lints edited files and reports lint errors back to Claude |
| grep MCP moved to `.mcp.json` (auto-approved) | ✅ Was misconfigured/inert; now connects and works |
| `.pnpm-store` gitignored | ✅ No effect on host/CI (the pnpm store is global there; this only matters inside a sandbox) |
| `supportedArchitectures` in `pnpm-workspace.yaml` | ⚠️ The one cost: a single install now also fetches the container's native binaries (~50 MB extra per install, CI included). No lockfile or repo-size change |

All positive except the extra binaries. Non–Claude-Code editors are unaffected apart from that.

### Context

Claude Code offers ways to cut down on permission prompts, but they leave the safety boundary in Claude's own hands:

- **Native sandbox mode** runs commands in a built-in OS sandbox — but when a sandboxed command fails (it needs network, wants to write outside the allowed paths, etc.) Claude can decide to re-run it _outside_ the sandbox. The model can step out of its own sandbox when it judges it necessary, so it's a convenience guardrail, not a hard boundary.
- **"Auto" approval** lets Claude judge which commands are safe and run them unprompted — making the model both the actor _and_ its own security reviewer. A model deciding what's safe to run is not a boundary you can rely on.

`sbx` (Docker Sandboxes) moves the boundary _outside_ the model: an OS-enforced container with its own filesystem, a default-deny network policy, and none of your host credentials or other projects (beyond anything you explicitly opt to pass in — e.g. a SonarCloud token for the mount). Because containment is enforced by the environment rather than the model's discretion, we run Claude with `--dangerously-skip-permissions` — zero prompts, uninterrupted flow — while the actual isolation is guaranteed by the sandbox, not by Claude's judgment. You get "approve nothing" speed with isolation that doesn't depend on the agent behaving.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

_"bypass permissions on" but that's OK because Claude is running in a restricted VM_
Screenshot 2026-06-17 at 12 30 19


michael-markevich (dismissed):
Looks good! I mostly focused on changes to sbx.sh / settings.json / .mcp.json.

edoardo (approved):

PR #292, opened 2026-06-23
chore: only use dynamic name fields for objects that support shortName

HendrikThePendric:
Implements [DHIS2-20103](https://dhis2.atlassian.net/browse/DHIS2-20103)

### Description

Only request `displayShortName` for objects that actually have a `shortName`. Objects without one (`programStage`, `option`, `optionSet`, `legendSet`, `legend`) now always use `displayName`, so their labels no longer render blank under the "short name" setting.

For the remaining objects that do have a `displayShortName` field we could just keep things as they were, because for the field is required. So no fallback is needed.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added (code comments)

---

[DHIS2-20103]: https://dhis2.atlassian.net/browse/DHIS2-20103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -13,13 +13,13 @@ importers:
     dependencies:
       '@dhis2/analytics':
         specifier: ^29.5.3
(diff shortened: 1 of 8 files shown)


edoardo (approved):
Nice to have this cleaned up!

PR #301, opened 2026-06-29
fix: improve layout resize handling [DHIS2-21641]

edoardo:
Implements [DHIS2-21641](https://dhis2.atlassian.net/browse/DHIS2-21641)

### Description

Improvements for the layout resize.
This has been simplified from the original spec.

See the ticket for a description of how it should work.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

New "Resize layout to fit" item in View menu:
Screenshot 2026-06-29 at 12 35 50

Active drop target style:
Screenshot 2026-06-29 at 12 39 07

Max height limit at 80vh, regardless of content:
Screenshot 2026-06-29 at 12 40 15

Auto resize when adding/removing chips, fit to content:
Screenshot 2026-06-29 at 15 44 04

Auto resize when adding/removing chips, with user-defined height axes are equally sized in PT:
Screenshot 2026-06-29 at 12 41 31

[DHIS2-21641]: https://dhis2.atlassian.net/browse/DHIS2-21641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/layout-panel/axis/axis.tsx
+++ b/src/components/layout-panel/axis/axis.tsx
@@ -42,9 +42,8 @@ export const Axis: FC<AxisProps> = ({ axisId, dimensionIds = EMPTY_ARRAY }) => {
         disabled,
     })
 
-    /* The collision detector resolves a drop over a populated axis to one of its
-     * chips, not to the axis container, so the container's own `isOver` only
-     * fires on an empty axis. Read the active drop target's axis instead — both
+    /*
+     * Read the active drop target's axis — both
      * chip sortables and the axis container carry it — so the whole axis
      * highlights whether the drop lands on a chip or the empty area. */
     const overData = over?.data.current
--- a/src/components/layout-panel/axis/axis.tsx
+++ b/src/components/layout-panel/axis/axis.tsx
@@ -30,38 +30,48 @@ export const Axis: FC<AxisProps> = ({ axisId, dimensionIds = EMPTY_ARRAY }) => {
         [axisId]
     )
 
-    const { active } = useDndContext()
+    const { active, over } = useDndContext()
     const disabled = useMemo(
         () => getActiveDragData(active)?.isLayoutBlocked ?? false,
         [active]
     )
 
-    const { setNodeRef, isOver } = useDroppable({
+    const { setNodeRef } = useDroppable({
         id: axisId,
         data: axisContainerData,
         disabled,
     })
 
+    /* The collision detector resolves a drop over a populated axis to one of its
+     * chips, not to the axis container, so the container's own `isOver` only
+     * fires on an empty axis. Read the active drop target's axis instead — both
+     * chip sortables and the axis container carry it — so the whole axis
+     * highlights whether the drop lands on a chip or the empty area. */
+    const overData = over?.data.current
+    const isActiveDropTarget =
+        overData !== undefined && 'axis' in overData && overData.axis === axisId
+
     return (
         <SortableContext id={axisId} items={dimensionIds}>
             <div
+                ref={setNodeRef}
                 className={cx(classes.container, {
                     [classes.columns]: axisId === 'columns',
                     [classes.rows]: axisId === 'rows',
                     [classes.filters]: axisId === 'filters',
+                    [classes.activeDropTarget]: isActiveDropTarget,
                 })}
                 data-test={`axis-${axisId}`}
                 aria-disabled={disabled || undefined}
             >
                 <div className={classes.label}>{getAxisName(axisId)}</div>
                 <div
-                    ref={setNodeRef}
                     className={classes.content}
                     data-test={`axis-content-${axisId}`}
                 >
                     {dimensions.length === 0 ? (
                         <div className={classes.emptyAxisInsertMarkerAnchor}>
-                            {isOver && (
+                            {isActiveDropTarget && (
                                 <span className={insertMarkerClasses.marker} />
                             )}
                         </div>
--- a/src/components/layout-panel/axis/axis.tsx
+++ b/src/components/layout-panel/axis/axis.tsx
@@ -69,21 +69,19 @@ export const Axis: FC<AxisProps> = ({ axisId, dimensionIds = EMPTY_ARRAY }) => {
                     className={classes.content}
                     data-test={`axis-content-${axisId}`}
                 >
-                    {dimensions.length === 0 ? (
+                    {dimensions.length === 0 && isActiveDropTarget && (
                         <div className={classes.emptyAxisInsertMarkerAnchor}>
-                            {isActiveDropTarget && (
-                                <span className={insertMarkerClasses.marker} />
-                            )}
+                            <span className={insertMarkerClasses.marker} />
                         </div>
-                    ) : (
+                    )}
+                    {dimensions.length > 0 &&
                         dimensions.map((dimension) => (
                             <Chip
                                 key={dimension.id}
                                 dimension={dimension}
                                 axisId={axisId}
                             />
-                        ))
-                    )}
+                        ))}
                 </div>
             </div>
         </SortableContext>
--- a/src/components/layout-panel/constants.ts
+++ b/src/components/layout-panel/constants.ts
@@ -1,0 +1,9 @@
+export const AXES_HEIGHT_STORAGE_KEY = 'dhis2.event-visualizer.axesHeight'
+
+/* Sentinel for "no user-set height": the panel fits its content. Stored in the
+ * UI slice alongside numeric user-set heights, so the height is
+ * `number | typeof LAYOUT_PANEL_HEIGHT_AUTO_FIT`. A named sentinel reads more
+ * clearly at call sites than a bare `null`. */
+export const LAYOUT_PANEL_HEIGHT_AUTO_FIT = 'AUTO_FIT'
+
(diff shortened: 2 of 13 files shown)


src/components/layout-panel/axis/axis.tsx
```diff
@@ -30,38 +30,48 @@ export const Axis: FC<AxisProps> = ({ axisId, dimensionIds = EMPTY_ARRAY }) => {
...
         id: axisId,
         data: axisContainerData,
         disabled,
     })
 
+    /* The collision detector resolves a drop over a populated axis to one of its
+     * chips, not to the axis container, so the container's own `isOver` only
+     * fires on an empty axis. Read the active drop target's axis instead — both
+     * chip sortables and the axis container carry it — so the whole axis
+     * highlights whether the drop lands on a chip or the empty area. */
```
HendrikThePendric: I am in two minds about this comment:
- On the one hand it was very helpful for this review, because it explains the reasoning for removing `isOver` and using `overData` instead (more about this in another comment below)
- On the other hand once `isOver` is removed it will perhaps just add confusion, because it's referencing a  a past state of the code that's not visible unless you use git blame

So perhaps in the long run we'd be better of removing it completely.

src/components/layout-panel/axis/axis.tsx
```diff
@@ -30,38 +30,48 @@ export const Axis: FC<AxisProps> = ({ axisId, dimensionIds = EMPTY_ARRAY }) => {
...
+    const overData = over?.data.current
+    const isActiveDropTarget =
+        overData !== undefined && 'axis' in overData && overData.axis === axisId
```
HendrikThePendric, line 51: I think I have a type guard available for identifying "over data".

src/components/layout-panel/axis/axis.tsx
```diff
@@ -30,38 +30,48 @@ export const Axis: FC<AxisProps> = ({ axisId, dimensionIds = EMPTY_ARRAY }) => {
...
                 <div className={classes.label}>{getAxisName(axisId)}</div>
                 <div
-                    ref={setNodeRef}
                     className={classes.content}
                     data-test={`axis-content-${axisId}`}
                 >
                     {dimensions.length === 0 ? (
                         <div className={classes.emptyAxisInsertMarkerAnchor}>
-                            {isOver && (
+                            {isActiveDropTarget && (
```
HendrikThePendric: Now this part I don't understand at all. My brain is telling me that this should be causing 2 insert markers to appear when the axis is populated, but I have inspected the DOM and my eyes have told me that this is not the case.

Perhaps I can explain what I _think_ would be going wrong, and then you can explain why it isn't:

1. The insert marker should show at the start/end of a chip, unless a the axis is empty and then the marker should be inserted directly into the axis
2. As the comment above states, we used to use `isOver` which only toggles true when the axis is empty, so then the insert marker would only get inserted into the axis when there are no chips
3. The `overData.axis === axisId` condition on the other hand returns `true` for both populated as well as empty axes, according to the comment and also judging by the fact that the inset shows correctly.
4. So as far as I can tell, that would mean that for populated axis we'd see both the insert marker in the axis and in the nearest chip.

As I mentioned, what I expect in point 4 is not actually happening in the DOM, but I really don't understand why that is. Can you explain it?

HendrikThePendric (commented):
I will continue reviewing tomorrow, but I already have a couple of questions, and am currently unsure if I should approve or request changes, so I'll just post a "first review" now.

src/components/layout-panel/axis/axis.tsx
HendrikThePendric: I completely missed the `dimensions.length === 0` condition on the container. This solves the mystery I now understand why this works. But I also think that my misunderstanding stems from the JSX being overly complex. This was something pre-existing, but let's tackle it now. We are in fact dealing with two separate conditions and JSX:
1. When `dimensions.length === 0 && isActiveDropTarget` we show the marker and its anchor
2. When `dimensions.length > 0` we map over the chips

HendrikThePendric (changes requested):
The `resetUiLayoutPanelHeightCounter` mechanism feels like a bit of a code smell to me. It models a one-shot "refit now" event as an ever-incrementing counter in the store, which then needs a ref and an effect that has to skip the first render to turn back into "fire once."

I think we can avoid it entirely by following the same pattern we already use for the sidebar width. The sidebar is slightly simpler, because the reset target value is static `SIDEBAR_DEFAULT_WIDTH`. The layout's reset target is DOM-computed, which is why a value couldn't be dispatched/used directly — but we can use `number | 'AUTO_FIT' instead: store the user-set height in Redux as a `number` or use 'AUTO_FIT' to mean "no user-set height, fit to content"), and make "Resize to fit" dispatch 'AUTO_FIT'. The consumer reacts to it by measuring and fitting, so the DOM computation stays where it belongs and only the intent crosses through the store. A named sentinel like 'AUTO_FIT' reads more clearly than null here, since it spells out what the state actually means.

That lets us drop the counter field/action/selector and the skip-first-render effect, and makes both resizable panels consistent. Bonus: while the height is 'AUTO_FIT' the panel already auto-tracks content, so reset is only meaningful when a custom height is set — we could disable the menu item in that case, like we do for the sidebar.

src/components/layout-panel/axis/axis.tsx
edoardo: Agreed.
Refactored the ternary operator into 2 conditional rendering blocks as described above.

src/components/layout-panel/axis/axis.tsx
HendrikThePendric: Looked into this and I don't have a type-guard for this

src/components/layout-panel/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const AXES_HEIGHT_STORAGE_KEY = 'dhis2.event-visualizer.axesHeight'
+
+/* Sentinel for "no user-set height": the panel fits its content. Stored in the
+ * UI slice alongside numeric user-set heights, so the height is
+ * `number | typeof LAYOUT_PANEL_HEIGHT_AUTO_FIT`. A named sentinel reads more
+ * clearly at call sites than a bare `null`. */
+export const LAYOUT_PANEL_HEIGHT_AUTO_FIT = 'AUTO_FIT'
```
HendrikThePendric: I don't think we should actually use a const for things like this in a TS codebase. If the type is like this we can safely use the string literal in the code.
```ts
type LayoutPanelHeight = number | 'AUTO_FIT`
```

src/components/layout-panel/use-resize-handle.ts
```diff
@@ -17,59 +20,52 @@ export const useResizeHandle = ({
...
-    const storedSizeRef = useRef<number | null>(
-        (() => {
-            const storedSize = localStorage.getItem(storageKey)
-
-            return storedSize ? Number(storedSize) : null
-        })()
-    )
+    /* Redux owns the user-set height. Mirror it into a ref so the callbacks
+     * below read the latest value without being re-created on every change. */
+    const storeHeightRef = useRef(storeHeight)
```
HendrikThePendric: Instead of working with a ref, you can use the `useStore()` hook. This is a stable ref that allows you to read state values in callbacks.

HendrikThePendric (changes requested):
Two minor comments

src/components/layout-panel/constants.ts
HendrikThePendric, line 1: Minor: I suspect that since we have a getter and setter for the local storage, that we can colocate the key with those in `src/components/layout-panel/local-storage.ts`

src/store/ui-slice.ts
```diff
@@ -98,6 +104,16 @@ export const uiSlice = createSlice({
         toggleUiLayoutPanelExpanded: (state) => {
             state.isLayoutPanelExpanded = !state.isLayoutPanelExpanded
         },
+        /* The user-set layout panel height, or AUTO_FIT for "no user-set height,
+         * fit to content". AUTO_FIT is how "Resize layout to fit" (and the
+         * resize handle's double-click) ask the panel to re-fit — the panel
+         * measures the content itself, since the fit height is DOM-computed. */
```
HendrikThePendric: Don't think we need the comment here TBH

HendrikThePendric (dismissed):
LGTM I left 2 minor comments, which you can either implement or leave

HendrikThePendric (approved):

src/store/ui-slice.ts
edoardo: Removed

src/components/layout-panel/constants.ts
edoardo: Done

src/components/layout-panel/use-resize-handle.ts
edoardo: Done.

src/components/layout-panel/constants.ts
edoardo: Done

src/components/layout-panel/axis/axis.tsx
edoardo: Rephrased the comment

PR #315, opened 2026-07-17
fix: preserve options selection when searching in optionSet modal (DHIS2-21831)

edoardo:
Implements [DHIS2-21831](https://dhis2.atlassian.net/browse/DHIS2-21831)

### Description

Selected options that didn't match the search term where removed on the right side of the Transfer.

In the optionSet modal, which uses the `Transfer` component, there was a bug when searching for options on the left side of the `Transfer`.
If there were previously selected on the right-end side, the ones that didn't match the search term where removed.
This didn't affect the Redux store, so all the selected options were correctly preserved. If you were re-opening the modal you would find all of them again.
The issue was internal to the component, caused by the options lookup using the search-narrowed data.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Known issues

I noticed when doing a search which does not match anything on the left side, but does match something on the right side, there is no text in the left side.
This is visible in the screenshots below.
The search for `char` matches the selected `Discharged` option, but nothing on the source side of the Transfer.
Normally when nothing is found a text is shown:
Screenshot 2026-07-17 at 12 30 37

I think the "Not found" message should show whenever a search does not have a match in the source side, regardless of what is present in the selected side. But this might be the way the `Transfer` works internally and nothing that can be controlled from the app.

---

### Screenshots

Before:
Screenshot 2026-07-16 at 15 59 30

After:
Screenshot 2026-07-17 at 12 27 15

[DHIS2-21831]: https://dhis2.atlassian.net/browse/DHIS2-21831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HendrikThePendric (changes requested):
I don't agree with this fix. It's hard to follow because the solution doesn't fit the problem very neatly. We're now introducing yet another data structure to populate one we already had. And we're populating it with all the fetched data, not just the selected data.

My suggestion is to implement something like this:
1. When the component mounts there may already be selected items, which can be sourced from the `optionSetMetadata`
2. After this, when `onChange` runs items may also be added or removed. In this case the new items must be present in `data`
3. So a solution that fits the problem better would then:
    - implements a `selectedOptionsLookupRef` that we'd write to when `optionSetMetadata` changes and when `onChange` fires
    - offers a way to lookup options from the `data` in a time-efficient way (not a nested iteration). I guess this fits nicely in the hook

HendrikThePendric (approved):

PR #316, opened 2026-07-21
fix: visualization name persistence [DHIS2-18726]

HendrikThePendric:
Implements [DHIS2-18726](https://dhis2.atlassian.net/browse/DHIS2-18726) and more

### Description

The reported bug — a visualization losing its name on save — was only the start. It turned out that `savedVis` (what the API returns) and `currentVis` (rebuilt from the UI config) differ in several ways, and those differences were making an unchanged visualization show as "Edited". This PR fixes the name bug as well as the "Edited" (state) bugs.

The `savedVis` and `currentVis` never match field-for-field, and there is no single shape to reduce both to: the app can't recreate the default values and computed fields the API adds, and stripping those off the saved object would throw away real data. So instead of striving for pure object equality, I implemented a custom comparator to detect changes.

Below I will list various field discrepancies in detail and explain how I dealt with them.

#### Name and description
When `currentVis` is built from `visUiConfig` or from `savedVis` it does not get these fields (they are not present in the config and `toCurrentVis` strips them). But when renaming the fields were being added, which was inconsistent behaviour.  And this was also causing the initial/reported bug: the name and description were being stripped after clicking update and the fallback text would show. 

Since name and description are saved directly and don't have a "dirty" state like the rest of the visualization fields, it makes sense to always exclude them from the `currentVis` and read from / write to the `savedVis`. This also means these fields need to be excluded when comparing `currentVis` and `savedVis`. This is not done explicitly, but since the `areVisualizationsEquivalent` function iterates over the fields of `currentVis`, they are skipped.

#### Fields ignored by the comparator
The following fields don't affect what gets saved, so a difference there does not have to be taken into account when comparing both objects:
- `dimensionType` / `valueType` — stripped before saving (we rename `PROGRAM_DATA_ELEMENT` to `DATA_ELEMENT` so without stripping these we'd see the comparator returning false positives)
- empty vs missing `items` — both mean "no items"
- `trackedEntityType` / `attributeDimensions` / `programDimensions` — derived from the layout, so any real change is already caught by comparing the axes; comparing them directly adds nothing. The two array-valued ones (`programDimensions`, `attributeDimensions`) can still differ in ordering between a loaded `savedVis` and a rebuilt `currentVis` — the app rebuilds them from the layout, while the backend returns its own order (only when one of those lists has more than one entry) — which a direct compare would misread as an edit. (`trackedEntityType` is a single `{id, name}`, so it just matches.)
- default-valued options — a default value is the same as an absent one (`getNonDefaultOptions`)

#### Legacy visualizations that use `ou` where the app now uses `enrollmentOu`
This is the one difference that does have a single correct form, here we just need to make sure that the normalized `savedVis` uses the same dimension ID as would be produced in the app when creating `currentVis` from `visUiConfig`. So this is handled in `normalizeApiSavedVisualization`, where the vis is also flagged as `legacy` so it isn't silently re-saved in the new format.

#### Additional explanation for `DEFAULT_OPTIONS`
While working on this PR it became apparent that there are some intricacies regarding various option defaults. These have been explained in the code with some additional comments, but I'll mention them here too:
- `digitGroupSeparator` effectively does not have a default value. This is because the system-setting value is not fixed, and there is no "use system default" option for it in the app. So the visualization's value for this field is seeded from the instance system setting at creation. So it is never ignored and always compared.
- `legend` (undefined = no legend; `DEFAULT_LEGEND_OPTION` is the seed used when it is turned on)
- `hideEmptyColumns` is commented out and I added some comments to explain why

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

[DHIS2-18726]: https://dhis2.atlassian.net/browse/DHIS2-18726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edoardo (dismissed):
Looks good.

edoardo (approved):

PR #331, opened 2026-08-04
fix: solve React error in condition modal Tabs [DHIS2-20947]

edoardo:
Implements [DHIS2-20947](https://dhis2.atlassian.net/browse/DHIS2-20947)

### Description

Opening the conditions modal for a data element on a non-repeatable stage (e.g. Child Programme → MCH Apgar Score) logged a React warning on first load:

> Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()? 
> Check the render method of Tabs.

`@dhis2/ui`'s `TabBar` renders an internal `Tabs` component that clones each direct child and injects a ref into it (used to call .focus() for keyboard navigation).
This works for `Tab` because it's a `forwardRef` component.
The disabled "Repeated events" tab was wrapped in a `Tooltip`, making `Tooltip` — a non-forwardRef function component — the direct child of `TabBar`. The injected ref hit the function component and triggered the warning.

The fix Introduced a small `forwardRef` wrapper (`RepeatableEventsTabTooltip`) that is now the direct child of `TabBar`. It forwards `TabBar`'s ref and the `Tooltip`'s anchor ref onto a single `<span>` element (via an `assignRef` helper), with the disabled `Tab` nested inside.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

React error in console for reference:

Screenshot 2026-08-04 at 15 48 20

[DHIS2-20947]: https://dhis2.atlassian.net/browse/DHIS2-20947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dimension-modal/conditions-modal-content/conditions-modal-content.tsx
+++ b/src/components/dimension-modal/conditions-modal-content/conditions-modal-content.tsx
@@ -3,7 +3,13 @@ import { Tooltip, TabBar, Tab } from '@dhis2/ui'
 import { useAppSelector, useProgramStageMetadataItem } from '@hooks'
 import { getVisUiConfigVisualizationType } from '@store/vis-ui-config-slice.js'
 import type { DimensionMetadataItem } from '@types'
-import { forwardRef, useState, type FC, type ReactNode, type Ref } from 'react'
+import {
+    forwardRef,
+    useState,
+    type FC,
+    type MutableRefObject,
+    type ReactNode,
+} from 'react'
 import { ConditionsTabContent } from './conditions-tab-content'
 import { RepeatedEventsTabContent } from './repeated-events-tab-content'
 import classes from './styles/conditions-modal-content.module.css'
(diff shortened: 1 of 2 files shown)


src/components/dimension-modal/conditions-modal-content/conditions-modal-content.tsx
```diff
@@ -2,15 +2,58 @@ import i18n from '@dhis2/d2-i18n'
...
 const TAB_CONDITIONS = 'CONDITIONS'
 const TAB_REPEATABLE_EVENTS = 'REPEATABLE_EVENTS'
 
+const assignRef = <T,>(ref: Ref<T>, node: T | null): void => {
+    if (typeof ref === 'function') {
+        ref(node)
+    } else if (ref) {
+        ;(ref as { current: T | null }).current = node
+    }
+}
```
HendrikThePendric: I'm finding this quite hard to read, mostly due to the type definitions. Would this work as well?


```suggestion
type MaybeElement = HTMLElement | null
type FunctionOrObjectRef = (MaybeElement) => void | { current: MaybeElement }
const assignRef = (ref: FunctionOrObjectRef, node: MaybeElement): void => {
    if (typeof ref === 'function') {
        ref(node)
    } else if (ref) {
        ref.current = node
    }
}
```

You'd also have to remove the `Ref` import. And I only typed this in the text-box, so it could contain errors and probably needs to be linted.

src/components/dimension-modal/conditions-modal-content/conditions-modal-content.tsx
```diff
@@ -2,15 +2,58 @@ import i18n from '@dhis2/d2-i18n'
...
+            {({ onMouseOver, onMouseOut, ref: tooltipRef }) => (
+                <span
+                    ref={(node) => {
+                        assignRef(tabBarRef, node)
+                        assignRef(tooltipRef, node)
+                    }}
+                    onMouseOver={onMouseOver}
+                    onMouseOut={onMouseOut}
+                    className={classes.tooltipReference}
+                >
```
HendrikThePendric: I think these suggestions are purely theoretical, because the disabled tab won't get receive focus, and therefore won't blur either. I wonder if we could just add
```
onFocus={onMouseOver}
onBlur={onMouseOut}
```

If this doesn't break anything it would help us get rid of the Sonar Qube error.

Alternatively, if it's possible to tell Sonar Qube to ignore these violations here, that would be good as well.

HendrikThePendric (commented):
I think this fix makes sense if it addresses the error. It is kind of silly though:
- The tab element needs a ref for keyboard nav
- This is causing issues on a disabled item that we want to wrap in a tooltip
- And these items are actually not accessible via keyboard nav 🙃 

Perhaps there are opportunities to simplify this a bit.

src/components/dimension-modal/conditions-modal-content/conditions-modal-content.tsx
HendrikThePendric: After closer inspection I think the cleanest way to deal with these is to simply accept the issues flagged up by Sonar Qube in its dashboard. The fix above is not great because then we attach focus handlers to a non-focusable element and we'd still have a Sonar Qube issue.

src/components/dimension-modal/conditions-modal-content/conditions-modal-content.tsx
HendrikThePendric: This function is now gone.

HendrikThePendric (approved):

PR #341, opened 2026-08-13
feat: implement error handling [DHIS2-20103]

HendrikThePendric:
Implements [DHIS2-20103](https://dhis2.atlassian.net/browse/DHIS2-20103)

### Description

This defines how the failures that actually come up are handled, and routes each one to the right place. It replaces the older  behaviour where a failure could leave a blank canvas or surface as a dismissible alert that's easy to miss.

Handling covers three kinds of failure:
1. Data errors — a failed analytics request or an empty response — show a message on the canvas with a Retry button, since these are sometimes transient or fixable by changing the selection. Known backend rejections (no access, bad request) show a specific message without Retry, because retrying won't help.
2. Visualisation-load errors (fetching the saved visualisation itself) follow the same idea: a network failure is retryable on the canvas, a data-processing failure is treated as fatal.
3. A render crash anywhere below us is caught by the error boundary the DHIS2 app shell already provides above our code, for both the app and the plugin, which shows a full reload screen. This is a deliberate simplification. Rather than keeping our own top-level fatal boundary, the shell's boundary is reused, since it catches the same render crashes and offers a reload. A parallel one of our own added complexity without  adding coverage. The canvas-level handling stays app-specific, so that part remains ours.

This simplified approach does not try to catch everything, and that's on purpose. An error boundary only catches crashes that happen while rendering. It cannot catch async errors — a rejected promise, or a throw inside an event handler or timer — unless they're  explicitly routed to it. The real async sources here (analytics and visualisation loading) are routed, so those are covered. A stray async error we didn't anticipate is not: it logs to the console and the app keeps running.

The previous approach did try to catch _everything_, with global window listeners that turned every uncaught async error into a fatal overlay. That was worse, not better. Some of the things that would trigger these listeners is harmless noise — browser extensions, third-party scripts, and the like — none of which actually breaks the app. Blanking a working screen over that is a false alarm, so those listeners were removed. The basic idea is that a boundary handles render crashes well, and reaching for a global net to cover the rest costs more than it's worth.

#### Note about the remount key

The canvas remount key is derived purely from props (visualisation plus dashboard filters), so it behaves the same in the store-less plugin as in the full app. Filters have been made part of the key on purpose: changing them changes the underlying result set, which invalidates the current pager, so the canvas has to remount and reset paging rather than refetch in place. Sorting and paging don't invalidate the pager that way, so they stay out of the key and refetch in place.

---

### Quality checklist

- [x] Dashboard tested N/A
- [x] Docs added N/A

---

### Screenshots

_TBD_

[DHIS2-20103]: https://dhis2.atlassian.net/browse/DHIS2-20103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/canvas/canvas.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <>
+            <PluginWrapper
+                isVisualizationLoading={isVisualizationLoading}
+                visualization={currentVis}
+                visualizationLoadError={visualizationLoadError ?? undefined}
+                onRetryLoad={onRetryLoad}
+                displayProperty={currentUser.settings.displayProperty}
+                onDataSorted={onDataSorted}
+            />
+            <InterpretationModal />
```
edoardo: This I'm a bit unsure about.
I thought `Canvas` is only for rendering the plugin and the canvas error.
The `InterpretationModal` seems to be "unrelated" to the `Canvas` as it's not technically part of it, but I guess because the `IntepretationModal` also uses the plugin and fetches the visualization it needs the same error handling.
Perhaps it's just the `Canvas` naming that confuses me.

edoardo (dismissed):
This looks good. I'm approving so you can merge before I'm back.

I added one comment which is not an issue per se.
If there is a good reasoning behind it we don't need to change anything.

src/components/canvas/canvas.tsx
HendrikThePendric: Good point. I agree that `InterpretationsModal` doesn't fit in the `Canvas`. I initially thought we would need to keep things this way because the early return to `StartScreen` without the `InterpretationsModal`. But I realised that the `InterpretationsModal` hides itself so it doesn't matter if we render it unconditionally. So I removed it in https://github.com/dhis2/event-visualizer-app/pull/341/commits/3737cd749ad49d1fd560ea131ca976c1e83c70b5

HendrikThePendric:
I will use my admin rights to merge this, since you already approved with one comment that I ended up implementing. For the record: I did wait for all checks to be green.

PR #352, opened 2026-08-25
feat: complete line list to pivot table layout conversion [DHIS2-21994]

HendrikThePendric:
Converting a line list to a pivot table kept every column on the columns axis, which a pivot table cannot render. Route each column by dimension type instead: org unit dimensions become columns, period dimensions become rows, and everything else is appended to the filters.

Restructure convertLayoutForVisType around the two conversion directions so each reads as its own spec: partition the layout into valid and discarded dimensions first, then hand the valid ones to either convertToLineList or convertToPivotTable.

Implements [DHIS2-XXXX](https://dhis2.atlassian.net/browse/DHIS2-XXXX)

### Description

_text_

---

### Quality checklist

- [ ] Dashboard tested
- [ ] Cypress and/or Jest tests added/updated
- [ ] Docs added
- [ ] d2-ci dependency replaced (requires <https://github.com/dhis2/analytics/pull/XXX>)

---

### ToDos

- [ ] _todo_

---

### Known issues

-   _issue_

---

### Screenshots

_supporting images_


src/modules/layout.ts
```diff
@@ -195,8 +195,21 @@ export const resolveLayoutContext = (
...
+}
+
+const getDimensionOrThrow = (
+    id: string,
+    getDimension: (id: string) => DimensionMetadataItem | undefined
+): DimensionMetadataItem => {
+    const dimension = getDimension(id)
+    if (!dimension) {
+        throw new Error(
+            `No metadata found for dimension "${id}" — cannot convert layout to a pivot table`
```
edoardo, line 209: This is fine for now since we only have LL and PT.
Since the function name is generic I'd expect the message would also need to be more generic (or interpolate the vis type) later on when we add more types and likely there will be more conversion rules.

src/modules/layout.ts
```diff
@@ -208,39 +221,52 @@ export const convertLayoutForVisType = ({
...
-            const targetAxis: Axis =
-                targetVisType === 'LINE_LIST' && sourceAxis === 'rows'
-                    ? 'columns'
-                    : sourceAxis
-            newLayout[targetAxis].push(dimensionId)
         }
     }
 
-    return { newLayout, discardedDimensionIds }
+    /* A line list has no rows in the layout, so there is no need to iterate
```
edoardo, line 267: This comment confused me at first.
I initially thought it referred to the case of `targetVisType === 'LINE_LIST'` but then realised is actually for  PT (and other) vis types, so the placement here is correct.

An observation about this file:
I see generic names for the functions are used but the logic after the return for LL is implicitly for PT only, which is fine at this stage.

src/modules/layout.ts
HendrikThePendric: Yeah this message is too specific. But this helper is removed in #354 so not an issue

src/modules/layout.ts
HendrikThePendric: Yeah we'll need to revisit this when we add more vis types. What we have is OK for now, but we know it'll get changed later.

